PR #1, opened 2018-08-17
Genrule setup

brandjon:
This is a proposal to have the custom initialization hook for genrules also apply to Starlark shell actions.

### DIFF
--- a/proposals/2018-08-17-genrule-setup-for-starlark.md
+++ b/proposals/2018-08-17-genrule-setup-for-starlark.md
@@ -1,11 +1,14 @@
-<nbsp>|<nbsp>
----|---
-status | draft
-created | 2018-08-17
-updated | 2018-08-17
-author(s) | [brandjon@](https://github.com/brandjon)
-reviewer(s) | Starlark core reviewers
-discussion thread | [#2](https://github.com/bazelbuild/starlark/issues/2)
+---
+title: Genrule-setup for Starlark
+status: Draft
+created: 2018-06-18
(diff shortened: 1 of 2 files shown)


proposals/2018-08-17-genrule-setup-for-starlark.md
```diff
@@ -0,0 +1,88 @@
+<nbsp>|<nbsp>
+---|---
+status | draft
+created | 2018-08-17
+updated | 2018-08-17
+author(s) | [brandjon@](https://github.com/brandjon)
+reviewer(s) | Starlark core reviewers
+discussion thread | [#2](https://github.com/bazelbuild/starlark/issues/2)
```
laurentlb: I suggest a header like this: https://github.com/bazelbuild/proposals/pull/6/files

I was told that `GitHub renders it specially, and we can easily parse it programmatically in the future.`

brandjon: I had trouble with github misrendering that header, but I can try again. The example you linked lacks fields for "updated" and "discussion thread", can I still include those here? I think they ought to be required in all design docs, github's parsing notwithstanding.

laurentlb: Yes, you can include those (the document from the link was discussed/approved before the new process).

brandjon: Updated. I presume the field order doesn't matter, or if it does my order's acceptable?

PR #11, opened 2018-10-03
ToolchainInfo Schema

brandjon:
This is a proposal to make `ToolchainInfo` resemble `Target` objects, by having its constructor take a sequence of provider instances, and retrieving provider instances by keying on the provider type.

I do not intend to implement this in the near future, but I suspect the design may be non-controversial.


brandjon:
Any blockers to merging as a draft? (Sections / formatting / missing content)

PR #15, opened 2018-10-29
Fix specification for string.capitalize()

laurentlb:
This was a mistake, as discussed on https://github.com/google/skylark/issues/140

string.capitalize uses uppercase only for the first character, unlike
string.title which does it for the first character of every word.


laurentlb:
cc @alandonovan

PR #34, opened 2019-02-21
Update hyperlink for java implementation

nguyentruongtho:
The old link is broken due to skylark -> starlark rename


laurentlb:
Thank you!

PR #50, opened 2019-06-06
add bazelci presubmit and shell script to setup workspace

Quarz0:
To add Starlark tests in the common test suite to bazel continuous integration.


.bazelci/presubmit.yml
```diff
@@ -0,0 +1,9 @@
+---
+platforms:
+  ubuntu1604:
+    shell_commands:
+      - sh setup_workspace.sh
+    build_targets:
+      - "//test_suite:all"
+    test_targets:
+      - "//test_suite:all"
```
laurentlb: I suggest `//...` for both build_targets and test_targets.
This will automatically pick up any new test we add.

setup_workspace.sh
```diff
@@ -0,0 +1,23 @@
+# This file is used for Bazel CI to install rust and setup the workspace.
+
+# install rust
+apt-get update && apt-get install curl -y && curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+# move workspace file to parent dir
+mv test_suite/WORKSPACE WORKSPACE
```
laurentlb: Moving the WORKSPACE file seems brittle. This can break absolute labels in the BUILD files.

You can put the WORKSPACE file at the root of the repository instead (we might need it in the future for other things).

setup_workspace.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# replace cargo in workspace with the absolute path $HOME/.cargo/bin/cargo
+sed -i -e 's/cargo/\$HOME\/.cargo\/bin\/cargo/g' WORKSPACE
+
+# create an empty build file
+touch BUILD
```
laurentlb: You shouldn't need this.

setup_workspace.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# modify the testdata_path
+cat > test_suite/testenv.py <<EOF
+STARLARK_BINARY_PATH = {
+    "java": "external/io_bazel/src/main/java/com/google/devtools/starlark/Starlark",
+    "go": "external/net_starlark_go/cmd/starlark/darwin_amd64_stripped/starlark",
+    "rust": "external/starlark-rust/target/debug/starlark-repl",
+}
+STARLARK_TESTDATA_PATH = "test_suite/"
+EOF
```
laurentlb: Why do you need to rewrite the file?
Oh, is it just because the TESTDATA_PATH changes when you move the WORKSPACE file? That's another reason to avoid doing it. :)

PR #53, opened 2019-06-24
Remove expected error msgs from code before executing it

Quarz0:
Remove error msgs from code before executing it and fix some test cases by adding Rust's error msgs.

When an error occurs in the Rust implementation, the source line is included in the error message.
Since we used regex to search for the error msg after `###`, many errors were masked because `### <error>` was already included as part of the source line in the error msg.


laurentlb:
I think the tests will pass if you rebase

Quarz0:
Done

Issue #57, opened 2019-06-26
spec: missing definition of string.elems


Quarz0:
string.elems is supported in the 3 implementations of Starlark and mentioned in the [go-spec](https://github.com/google/starlark-go/blob/master/doc/spec.md#stringelems).

laurentlb:
We should update the spec

PR #62, opened 2019-07-02
Add string.elems

Quarz0:
#57

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -299,6 +299,7 @@ Strings have several built-in methods:
 
 * [`capitalize`](#string·capitalize)
 * [`count`](#string·count)
+* [`elems`](#string·elems)
 * [`endswith`](#string·endswith)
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -2788,6 +2789,16 @@ They are interpreted according to Starlark's [indexing conventions](#indexing).
 "hello, world!".count("o", 7, 12)       # 1  (in "world")
 ```
 
+<a id='string·elems'></a>
+### string·elems
+
+`S.elems()` returns a list containing successive
+1-byte substrings of S.
```
laurentlb: Replace list with iterable value.

Quarz0: But the Java implementation returns a list

laurentlb: And the Go implementation doesn't. Since lists are iterable, it's correct for both implementations to say we return an iterable.

Issue #56, opened 2019-06-26
spec: no mention of optional parameters start, end in string.startswith/endswith


Quarz0:
The [go-spec](https://github.com/google/starlark-go/blob/master/doc/spec.md#stringstartswith) does mention the start, end arguments. And both the Java and Go implementations support these optional arguments. Only the rust implementation is different where it only accepts the parameter suffix, which must be a string.

laurentlb:
We should update the spec (to match the Java/Go implementations).

PR #63, opened 2019-07-02
Update spec of string.startswith/endswith

Quarz0:
#56

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2791,10 +2791,20 @@ They are interpreted according to Starlark's [indexing conventions](#indexing).
 <a id='string·endswith'></a>
 ### string·endswith
 
-`S.endswith(suffix)` reports whether the string S has the specified suffix.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -2791,10 +2791,19 @@ They are interpreted according to Starlark's [indexing conventions](#indexing).
 <a id='string·endswith'></a>
 ### string·endswith
 
-`S.endswith(suffix)` reports whether the string S has the specified suffix.
+`S.endswith(suffix[, start[, end]])` reports whether the string 
+`S[start:end]` has the specified suffix.
 
 ```python
 "filename.sky".endswith(".sky")         # True
+"filename.sky".endswith(".sky", 9, 12)  # False
```
laurentlb: Add an exemple with start/end where it returns True. This can help understand how it works.

PR #68, opened 2019-07-16
Remove `not` from binary operators in the spec

vladmos:
`not` is not a binary operator and shouldn't be mentioned in their precedence order.


vladmos:
A similar PR to starlark-in-go: https://github.com/google/starlark-go/pull/227

PR #75, opened 2019-07-31
Add benchmark test runner

Quarz0:
Sample run:

```bash
$ bazel test //test_suite:benchmark_test_testdata/benchmark/empty_star --test_output=all
INFO: Analyzed target //test_suite:benchmark_test_testdata/benchmark/empty_star (3 packages loaded, 193 targets configured).
INFO: Found 1 test target...
Target //test_suite:benchmark_test_testdata/benchmark/empty_star up-to-date:
  bazel-bin/test_suite/benchmark_test_testdata/benchmark/empty_star
INFO: Elapsed time: 1.884s, Critical Path: 0.03s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
PASSED: //test_suite:benchmark_test_testdata/benchmark/empty_star (see /private/var/tmp/_bazel_marwan/36607e4d48395b98a846ca0c58296c3f/execroot/starlark_test_suite/bazel-out/darwin-fastbuild/testlogs/test_suite/benchmark_test_testdata/benchmark/empty_star/test.log)
INFO: From Testing //test_suite:benchmark_test_testdata/benchmark/empty_star
==================== Test output for //test_suite:benchmark_test_testdata/benchmark/empty_star:
.
----------------------------------------------------------------------
Ran 1 test in 0.983s

OK
=== testdata/benchmark/empty.star ===
         Time (MS)  Memory (KB)
java       1033.98     65060.00
go            8.79      2872.00
rust          6.55      2636.00
python       44.43      3932.00
================================================================================
//test_suite:benchmark_test_testdata/benchmark/empty_star       (cached) PASSED in 1.4s

Executed 0 out of 1 test: 1 test passes.
INFO: Build completed successfully, 1 total action
```

In addition to time and memory, it is also possible to obtain more info from [resource usage](https://docs.python.org/2/library/resource.html#resource.getrusage).


laurentlb:
cc @meisterT

test_suite/benchmark_test.py
```diff
@@ -0,0 +1,91 @@
...
+    def memory(self):
+      if sys.platform == 'darwin':
+        return "{0:.2f}".format(self.ru.ru_maxrss / 1024)
+      return "{0:.2f}".format(self.ru.ru_maxrss)
+
+
+  def benchmark(self, f, binary):
+    """Execute Starlark file, return stderr."""
+    proc = subprocess.Popen(
+        [glob.glob(testenv.STARLARK_BINARY_PATH[binary])[0], f],
```
meisterT: What type of benchmarks do you have in mind?
If these are micro-benchmarks, then the JVM overhead will dominate the Starlark runtime performance.

test_suite/benchmark_test.py
```diff
@@ -0,0 +1,91 @@
...
+            print(col, end='')
+        print()
+
+  def testFile(self):
+    print("===", test_file, "===")
+    f = os.path.join(testenv.STARLARK_TESTDATA_PATH, test_file)
+
+    results = []
+    for binary in testenv.STARLARK_BINARY_PATH:
+      result_, out = self.benchmark(f, binary)
```
meisterT: consider doing more than one run per binary to eliminate noise.

meisterT: for this you might find https://github.com/bazelbuild/bazel-bench/blob/master/utils/values.py useful /cc @joeleba

test_suite/benchmark_test.py
Quarz0: Not micro-benchmarks, but operations with a large number of iterations (10^8 for example) known complexity (sorting, hashing, string algorithms, ...). I know that JVM has a large overhead, but with such benchmarks, we can still get useful info. For example, having benchmarks for operations than should run in O(n) with 10^8 iterations, or O(n^2) with 10^4 iterations and so on, and comparing benchmarks with the same complexity with each other can help identify if some slower operation has an area for improvement. The same goes for comparing memory consumption between implementations. Values of time and memory may not have a meaning by themselves but comparing those values with other benchmarks may do...

meisterT: I discussed with Laurent offline today. If in the benchmarks, you make sure that you run enough iterations so that the JVM startup doesn't dominate the overall runtime of the test (say a couple of seconds), then it's fine.

Let me know if you already have a list of benchmarks in mind.

Quarz0: I added some benchmarks for the basic functions in Starlark. Please let me know what you think.
Other benchmarks I have in mind are, as I previously mentioned, some general algorithms and operations on data structures. Examples are benchmarks for graph algorithms, binary search, heaps, trees, tries, etc. I think they can be useful where if there are some performance issues in Starlark then these benchmarks would help detect them by producing results contradicting the expected complexities of these benchmarks.

meisterT: I wonder whether those general algorithms are a good indicator of starlark performance when it comes to its use in bazel. But it's a starting point, so feel free to go ahead.

meisterT:
I tried to run all benchmarks and I mostly get timeouts (on a relatively beefy machine) and the rust implementation doesn't work at all (so I commented it out).

Quarz0:
> I tried to run all benchmarks and I mostly get timeouts (on a relatively beefy machine) and the rust implementation doesn't work at all (so I commented it out).

Rust does indeed timeout in most of the tests.
However, it's strange that you got timeouts with other implementations. The results I got on my machine are close to those on [buildkite](https://buildkite.com/bazel/starlark/builds/98#f2273b68-8550-43e0-8e3c-7e3620d9fbab) (a bit faster on my machine) where you can see there are very few timeouts. Also, when writing these benchmarks I tried to set the number of iterations as high as possible and at the same time to avoid timeouts so the results are comparable.

meisterT:
> > I tried to run all benchmarks and I mostly get timeouts (on a relatively beefy machine) and the rust implementation doesn't work at all (so I commented it out).
> 
> Rust does indeed timeout in most of the tests.

Sorry, I was not precise enough there. I meant: the rust implementation is not working for me since I got compile errors, so I just commented it out.


> However, it's strange that you got timeouts with other implementations. The results I got on my machine are close to those on [buildkite](https://buildkite.com/bazel/starlark/builds/98#f2273b68-8550-43e0-8e3c-7e3620d9fbab) (a bit faster on my machine) where you can see there are very few timeouts. Also, when writing these benchmarks I tried to set the number of iterations as high as possible and at the same time to avoid timeouts so that the results are comparable.

We should make sure that the tests are configured in a way that they do not run into timeouts on buildkite.

Looking at the buildkite run I have one more request: please order the results in a fixed way (e.g. alphabetical or ordered by time desc).

Quarz0:
> Looking at the buildkite run I have one more request: please order the results in a fixed way (e.g. alphabetical or ordered by time desc).

I think Bazel somehow decides the order by which to run the tests. I couldn't actually find a way to force a specific order through Bazel.

laurentlb:
cc @alandonovan

alandonovan:
Interesting discussion on JMH vs Caliper: 
https://groups.google.com/forum/#!msg/mechanical-sympathy/m4opvy4xq3U/7lY8x8SvHgwJ
Many of the concerns he raises are rather low-level and would not apply to a call into the interpreter, but it shows I think an attention to detail.

Quarz0:
Thanks for the explanation, Alan. However, there are some things that I do not quite understand.
First, does this design mean that we should implement a new Stalark data type (B) that has fields N, star_timer(), stop_timer(), ...? And is this type going to be a standard Starlark data type?
Also, it's not clear to me how the examples you mentioned relate to the benchmark packages (Go's bench_test, Caliper, JMH).
Meaning that if we are going to move the time measurement using the timer of (b) inside the benchmark_* functions, then what is the purpose of those benchmark packages?
If you could perhaps provide a bit more detailed example starting with how the benchmark test is executed and how the results/output is to be displayed, that would be great and I think would make things more clear to me.
Thanks!

laurentlb:
Since the time is limited, I suggest you skip the new type for now. It can be added later.
(Alan's suggestion involves a new data type, to be declared in the benchmark binary, but it doesn't have to be in the language spec)

I think the simplest is to have:

Input file
```python
def benchmark_replace():
  "abc".replace("a", "b")
```

The interpreter parses the file and looks for all the `benchmark_` functions. It executes each function many times and reports the statistics.

meisterT:
What's the state of this?

alandonovan:
146cd2cecae52ebad6664046fd1430c2a23b55e1 implemented the approach described in https://github.com/bazelbuild/starlark/pull/75#pullrequestreview-275604129. Closing this PR.

PR #80, opened 2019-08-04
Fix a broken link for qri.io’s usage page

jornh:


jornh:
@googlebot I signed it!

PR #83, opened 2019-08-26
proposal: JSON Module

jmillikin-stripe:
This document proposes an API for encoding and decoding JSON from Starlark. If accepted, implementations of Starlark that implement JSON support would be expected to implement this API. The goal is to allow users to write JSON code that behaves consistently across Starlark implementations.

### DIFF
--- a/proposals/2019-08-26-json-module.md
+++ b/proposals/2019-08-26-json-module.md
@@ -7,7 +7,7 @@ authors:
   - [jmillikin](https://john-millikin.com)
 reviewers:
   - Starlark core reviewers
-discussion thread: [#](https://github.com/bazelbuild/starlark/issues/)
+discussion thread: [#83](https://github.com/bazelbuild/starlark/pull/83)
 ---
 
 # JSON Module
--- a/proposals/2019-08-26-json-module.md
+++ b/proposals/2019-08-26-json-module.md
@@ -1,0 +1,99 @@
+---
+title: JSON Module
+status: Draft
+created: 2019-08-26
+updated: 2019-08-26
+authors:
+  - [jmillikin](https://john-millikin.com)
+reviewers:
+  - Starlark core reviewers
+discussion thread: [#](https://github.com/bazelbuild/starlark/issues/)
+---
+
+# JSON Module
+
+## Abstract
+
+This document proposes an API for encoding and decoding JSON from Starlark. If accepted, implementations of Starlark
+that implement JSON support would be expected to implement this API. The goal is to allow users to write JSON code that
+behaves consistently across Starlark implementations.
+
+## Background
+
+JSON is a popular syntax for representing basic data structures. Partial support for parsing or serializing JSON from
+Starlark has been independently requested and/or implemented several times:
+
+* Bazel's [`struct`](https://docs.bazel.build/versions/master/skylark/lib/struct.html) type has a `to_json()` method
+  that can generate JSON documents. Issue [bazelbuild/bazel#7879](https://github.com/bazelbuild/bazel/issues/7879)
+  requests additional control over the behavior of this method (i.e. optional whitespace).
+
+* [bazelbuild/bazel#3732](https://github.com/bazelbuild/bazel/issues/3732) requests support for parsing JSON from
+  within a Bazel repository rule, for use with introspecting the state of external tools.
+
+* [`json_parser.bzl`](https://github.com/erickj/bazel_json) is a JSON parser implemented entirely within Starlark.
+  Limitations of the Starlark language prevent this parser from being recommended for production use.
+
+* [google/starlark-go#179](https://github.com/google/starlark-go/pull/179) proposes a JSON module for the Go
+  implementation of starlark.
+
+* [Skycfg](https://github.com/stripe/skycfg) has a `json.marshal()` method that can generate JSON documents, for use
+  with config formats based on JSON syntax.
+
+JSON implementations in the broader community have wildly variant APIs. It is likely that ad-hoc JSON extensions to
+Starlark will have different APIs between Starlark implementations.
+
+## Proposed API
+
+The JSON module is a value named `json` in the global namespace. Its API comprises the following functions. Starlark
+implementations are not required to support the entire API, but should avoid extending the API with non-standard
+functions or parameters.
+
+JSON implementations should comply with the format documented as
+[ECMA-404](https://www.ecma-international.org/publications/standards/Ecma-404.htm) and
+[RFC 8259](https://www.rfc-editor.org/rfc/rfc8259.html), which supercedes earlier drafts of JSON. In particular, the
+only permitted character encoding for contemporary JSON is UTF-8.
+
+To maintain compatibility with existing callers, new required parameters should not be added to these functions. New
+optional parameters should be defined using keyword-only parameters
+([PEP-3102](https://www.python.org/dev/peps/pep-3102/)).
+
+### json.decode()
+
+The `json.decode()` function decodes a JSON document into a Starlark value.
+
+```python
+def json.decode(data):
+```
+
+Type conversions are:
+* JSON arrays are decoded to Starlark `list` values.
+* JSON objects are decoded to Starlark `dict` values. Key are in the same order as the input data.
+* JSON `true`, `false`, and `null` literals are decoded to Starlark `True`, `False`, and `None` respectively.
+* JSON strings are decoded to Starlark `string` values.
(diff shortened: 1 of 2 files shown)


proposals/2019-08-26-json-module.md
```diff
@@ -0,0 +1,99 @@
...
+To maintain compatibility with existing callers, new required parameters should not be added to these functions. New
+optional parameters should be defined using keyword-only parameters
+([PEP-3102](https://www.python.org/dev/peps/pep-3102/)).
+
+### json.decode()
+
+The `json.decode()` function decodes a JSON document into a Starlark value.
+
+```python
+def json.decode(data):
```
JacobHayes: Very cool! Is there any reason we wouldn't follow the [python json interface](https://docs.python.org/3/library/json.html), perhaps only the string `loads`/`dumps` versions? They might have a few extra flags, but 99% of the uses would suffice with the flags already included in this spec.

jmillikin-stripe: I'm fine with following the Python API here. Historically Bazel hasn't worried too much about cross-compatibility with the Python stdlib, and I thought `loads` / `dumps` is a little opaque to people who don't use Python regularly, but I'd accept the functionality under ~any name.

bpowers:
@laurentlb do you (or anyone else) have concerns with this? 

I'd love to see this merged so that we could move towards an implementation.  A tool I'm working on needs JSON, and I've ended up copy-pasting files from this PR as there isn't a standard here yet: https://github.com/google/starlark-go/pull/179

proposals/2019-08-26-json-module.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+Type conversions are:
+* JSON arrays are decoded to Starlark `list` values.
+* JSON objects are decoded to Starlark `dict` values. Keys are in the same order as the input data.
+* JSON `true`, `false`, and `null` literals are decoded to Starlark `True`, `False`, and `None` respectively.
+* JSON strings are decoded to Starlark `string` values.
+* JSON numbers with no fractional component are decoded to Starlark `int` values. Starlark implementations without
+  arbitrary-precision integers should reject numbers that exceed their supported range.
+* JSON numbers with a fractional component may be decoded to an arbitrary-precision or floating-point value, if
```
ndmitchell, line 75: Is it conforming for implementations without floating-point values to keep them as a string? Or do they have to error?

proposals/2019-08-26-json-module.md
```diff
@@ -0,0 +1,99 @@
...
+  supported by the current Starlark implementation.
+  * Starlark implementations without arbitrary-precision numeric values should reject numbers that exceed their
```
ndmitchell, line 77: What if they are within the bounds, but exceed the precision to represent them unambiguously?

proposals/2019-08-26-json-module.md
jmillikin-stripe: An implicit type conversion of float to string would be very surprising to me as a user, so I think I'd prefer the implementation to error if it can't accept a given input value.

proposals/2019-08-26-json-module.md
jmillikin-stripe: I guess "range" is a little wrong here, but I mean numbers that the implementation can't represent. In the trivial case, a value `0.00[...]01` with sufficient zeros to overflow a 64-bit IEEE754 would be rejected by most implementations.

I hope that issues of floating point are mostly academic because most Starlark implementations don't allow floats by default, and the primary client codebase (Bazel extensions) doesn't allow floats at all.

jmillikin-stripe:
> What is the diff to the spec? Does this proposal go as a brand new file in the repo? Assuming its accepted, what diff would we apply to the spec?

No idea! Happy to format this in whatever way seems reasonable to you. It doesn't seem like there's much prior art here, so I formatted it as an RFC-type file.

> What is a module? How do you in a standards conforming way import the "json" module? I couldn't see any existing modules in the spec (maybe I missed them?). Or maybe you mean json is an always available name with some fields on it?

Bazel defines a list of "global modules" (https://docs.bazel.build/versions/2.0.0/skylark/lib/skylark-overview.html#global-modules) but doesn't go into detail about what exactly a "module" is at the language level. For this proposal I had imagined a new global symbol named `json` that would be present if the current implementation supported JSON. The presence of `json` is optional, but if present it should conform to this API.

> If so, it feels a bit weird to introduce a whole new mechanism not present elsewhere in the spec for just two functions. Or maybe you want to use this mechanism more widely?

If this proposal is successful + accepted then I would like to propose more modules, specifically `proto` (for protobufs as used in Skycfg) and `yaml` (for use in Bazel, parsing lock files from the JavaScript ecosystem).

ndmitchell:
> Bazel defines a list of "global modules"

But Starlark doesn't. Maybe we need a corresponding piece of the Starlark spec that says what a "module" is, whether it needs loading, how it behaves, whether it's a first-class value etc.

alandonovan:
Hi, back from vacation. Sorry this proposal has been neglected and that this has process not been a model of good open-source stewardship.

Regarding the the project: Starlark/Java was until very recently highly entangled with Bazel, though  after a year of work nearly all its production dependencies on "Bazel proper" are now severed. However, the tests are still entangled, and there are numerous subtle constraints imposed on Starlark by Bazel that I plan to break soon to unblock significant optimizations (e.g. flat environments, byte code compilation). In addition we plan to move the Java classes to a different package (java.starlark.net) to emphasize the boundary between components, though we have no plans yet to host them in a separate repository; that would of course require the API to be refined, stabilized, and frozen. I believe the separation is in the long term interests of the Bazel project and its users, especially if Starlark becomes a widely used and more familiar language for configuration, but for the Bazel team it does pose immediate costs (in ceded control), and the team has yet to make any public commitments about Starlark.

Regarding JSON: In June I committed https://github.com/google/starlark-go/blob/master/starlarkjson/json.go#L26, a JSON module I wrote some years ago for the Go implementation of Starlark (go.starlark.net). I believe its interface addresses all the needs met by the one proposed here, and indeed was influenced by it. It is not too late to change it, so if you see any shortcomings of deficiencies, please let me know. Otherwise I propose that we port it directly to Java.

Regarding modules: Starlark doesn't yet have a standard library of packages, though there is clearly a recurring need for things like json, time, http, and html.  (Brendan @b5 at https://github.com/qri-io/starlib has built many good candidates for standardization.)  We don't even have a clear decision yet on how such packages should be imported. Sometimes we use load(), other times the names are predeclared (as I did with 'json' in the Go Starlark REPL, though I don't think it's very satisfactory).  Personally I would like to extend load so that modules are first-class, e.g. something like `load("foo.bzl"); foo.F()`, similar to Python's import statement, as many users resort to doing this by hand using structs like `os = struct(mkdir = _mkdir, ...)`.

Regarding proto: I have ported the Go implementation of the Starlark proto module so that it depends only on open-source proto reflection (which is new), so I should be able to open-source it. However, proto support for Starlark/Java will be tricky, because it has no means of representing byte strings, floating-point numbers, or integers outside the signed 32-bit range, and each of those poses surprisingly thorny problems.

BTW: Nick: I enjoyed both of your papers on build systems!

alandonovan:
I just committed https://github.com/bazelbuild/bazel/commit/6e47a409914d9f2ccb169482252f142c9fb503b8, which adds a `json` module to Starlark/Java. Please take a close look at the Starlark and Java interfaces, and try it out, and comment here if notice any problems.

Thanks to @brandjon for many very careful rounds of review.

laurentlb:
What's the use-case for the `prefix` argument in `json.indent`?

alandonovan:
> What's the use-case for the prefix argument in json.indent?

Perhaps the caller wants to insert the result inside some other piece of text, or JSON header/footer, that imposes its own indentation.

alandonovan:
Both the Java and Go implementations now have compatible implementations of the json module. Closing.

jmillikin-stripe:
Since it's been implemented, shouldn't this PR be merged rather than closed? Optionally with updates to reflect whatever common API was decided upon by the Java/Go impl devs.

alandonovan:
Er, yes, sorry about that. I just noticed that I also have a bunch of old pending comments on your doc from months ago that some how never got flushed out; I'll do that now.

The Java implementation is here:
https://github.com/bazelbuild/bazel/blob/267bdaab2f0596b65b03c06c82e3f91d957fec3b/src/main/java/net/starlark/java/lib/json/Json.java
and it should match the behavior of the Go implementation here:
https://github.com/google/starlark-go/blob/master/starlarkjson/json.go
at least as far as the Starlark caller is concerned.

PR #87, opened 2019-10-30
Fix a grammar error

junhan-z:
"it also"


junhan-z:
@googlebot I signed it!

PR #107, opened 2020-09-24
Update spec from google/starlark-go

laurentlb:
- document call argument evaluation order
- load statements create local bindings 
- Rename "universe" to "predeclared"
- hash: allow only strings 
- Add TOC


spec.md
```diff
@@ -463,8 +463,7 @@ values include `None`, Booleans, numbers, and strings, and tuples
 composed from hashable values.  Most mutable values, such as lists,
 and dictionaries, are not hashable, unless they are frozen.
 Attempting to use a non-hashable value as a key in a dictionary
-results in a dynamic error, as does passing one to the built-in
-`hash` function.
+results in a dynamic error.
```
alandonovan, line 466: I think the old text was correct (does that mean this change needs to flow in the other direction?).

Both `hash` implementations accept only string, not list, say.

spec.md
```diff
@@ -463,8 +463,7 @@ values include `None`, Booleans, numbers, and strings, and tuples
 composed from hashable values.  Most mutable values, such as lists,
```
alandonovan, line 463: I noticed that the TOC that exists in Go spec has vanished from bazelbuild. Could you restore it?

alandonovan (approved):

spec.md
laurentlb: I think that's correct.

The old text said it's an error to pass a non-hashable value to hash. This might be technically true, but that's irrelevant since hash accepts only strings. No need to mention `hash` here.

spec.md
laurentlb: Done

spec.md
alandonovan: Oh, of course.

PR #108, opened 2020-09-24
Spec: document string literals

laurentlb:
Copied from https://github.com/google/starlark-go/pull/265
I omitted the implementation notes.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -299,7 +299,123 @@ octal_digit   = '0' … '7' .
 hex_digit     = '0' … '9' | 'A' … 'F' | 'a' … 'f' .
 ```
 
-TODO: define string_lit, indent, outdent, semicolon, newline, eof
+### String literals
+
+A Starlark string literal denotes a string value.
+In its simplest form, it consists of the desired text
+surrounded by matching single- or double-quotation marks:
+
+```python
+"abc"
+'abc'
+```
+
+Literal occurrences of the chosen quotation mark character must be
+escaped by a preceding backslash. So, if a string contains several
+of one kind of quotation mark, it may be convenient to quote the string
+using the other kind, as in these examples:
+
+```python
+'Have you read "To Kill a Mockingbird?"'
+"Yes, it's a classic."
+"Have you read \"To Kill a Mockingbird?\""
+'Yes, it\'s a classic.'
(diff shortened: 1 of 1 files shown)


laurentlb:
Note that it doesn't exactly match the Java implementation, e.g. we reject \a.
I assume you'd rather update the implementation?

spec.md
```diff
@@ -299,7 +299,123 @@ octal_digit   = '0' … '7' .
...
+
+An *escaped newline*---that is, a backslash at the end of a line---is ignored,
+allowing a long string to be split across multiple lines of the source file.
+
+```python
+"abc\
+def"			# "abcdef"
+```
+
+An *octal escape* encodes a single byte using its octal value.
```
alandonovan, line 354: (The reference to byte here isn't really appropriate for the Java implementation.)

spec.md
```diff
@@ -299,7 +299,123 @@ octal_digit   = '0' … '7' .
...
+It is error if the value is greater than decimal 255.
+
+```python
+'\0'			# "\x00"  a string containing a single NUL byte
+'\12'			# "\n"    octal 12 = decimal 10
+'\101-\132'		# "A-Z"
+'\119'			# "\t9"   = "\11" + "9"
+```
+
+A *hex escape* encodes a single byte using its hexadecimal value.
```
alandonovan, line 365: (and hex escapes aren't implemented in Java)

alandonovan (approved):

alandonovan:
> Note that it doesn't exactly match the Java implementation, e.g. we reject \a.
> I assume you'd rather update the implementation?

Yes, useless though it may be, every language supports \a thus so should we.

spec.md
laurentlb: The rest of the spec mentions that a "string represents an immutable sequence of bytes."
I think it's correct that an octal escape encodes a byte.

Do you have another phrasing to suggest?

spec.md
laurentlb: Removing

spec.md
alandonovan: This seems fine for now. Correctly documenting how the Java implementation differs is a complex and subtle project.

PR #118, opened 2020-10-14
spec: fix typo in raw string example

alandonovan:


alandonovan:
See https://github.com/google/starlark-go/pull/311 for go.starlark.net change.

brandjon (approved):

PR #119, opened 2020-10-15
spec: add floating-point numbers

alandonovan:


spec.md
```diff
@@ -1722,8 +1845,15 @@ String interpolation
    string % any                 # see String Interpolation
 ```
 
-The operands of the arithmetic operators `+`, `-`, `*`, `//`,`%`, `&`,
-`|` and `^` must both be `int`. The type of the result has type `int`.
+The operands of the arithmetic operators `+`, `-`, `*`, `//`, and `%`,
+must both be numbers (`int` or `float`) but need not have the same type.
```
brandjon, line 1849: Do we mention somewhere in the spec the difference with Python that booleans are not numbers?

spec.md
```diff
@@ -1696,23 +1817,25 @@ The following table summarizes the binary arithmetic operations
...
+   number / number              # real division (result is always a float)
    number // number             # floored division
    number % number              # remainder of floored division
-   number ^ number              # bitwise XOR
-   number & number              # bitwise AND
-   number | number              # bitwise OR
-   number << number             # bitwise left shift
-   number >> number             # bitwise right shift
+
+Bitwise operations:
```
brandjon, line 1828: Add `(int only)`?

spec.md
```diff
@@ -1827,6 +1957,12 @@ operand types are valid and how to convert the operand `x` to a string:
...
 d       number          signed integer decimal
+o       number          signed octal
+e       number          float exponential format, lowercase (1.230000e+12)
+E       number          float exponential format, uppercase (1.230000E+12)
+f       number          float decimal format                (1230000000000.000000)
+g       number          compact format, lowercase           (0.0, 1.1, 1200, 1e+45, 1.2e+12) 
+G       number          compact format, lowercase           (0.0, 1.1, 1200, 1e+45, 1.2E+12)
 ```
 
 It is an error if the argument does not have the type required by the
```
brandjon, line 1968: Except ints can be promoted to floats and floats can be truncated to ints.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
 ```
 
 
+### Floating-point numbers
+
+The Starlark floating-point data type represents an IEEE 754
+double-precision floating-point number.
+Its [type](#type) is `"float"`.
+
+Arithmetic on floats using the `+`, `-`, `*`, `/`, `//`, and `%`
```
brandjon, line 538: Mention that arithmetic on floats can overflow the magnitude, resulting in an error. (Note that Starlark's lack of error handling makes it pretty unsuitable for intensive floating point computations, even if we didn't have the NaN issues.)

spec.md
```diff
@@ -2606,6 +2742,24 @@ enumerate(["zero", "one", "two"])               # [(0, "zero"), (1, "one"), (2,
...
+`float(x)` interprets its argument as a floating-point number.
+
+If x is a `float`, the result is x.
+
+If x is an `int`, the result is the floating-point value nearest x.
+The call fails if x is too large to represent as a `float`.
+
+If x is a `bool`, the result is `1.0` for `True` and `0.0` for `False`.
+
+If x is a string, the string is interpreted as a floating-point literal.
```
brandjon, line 2756: Ah, so this is why `float(<crazy large int>)` fails with overflow but `float(<string repr of crazy large int>)` succeeds as inf. In that case, seems like we can't make float's behavior consistent unless we also make it an error to write crazy large float literals (which we should also specify the behavior of in this doc). I think I'd be up for making all those cases fail.

spec.md
```diff
@@ -2606,6 +2742,24 @@ enumerate(["zero", "one", "two"])               # [(0, "zero"), (1, "one"), (2,
...
+The function also recognizes the names `Inf` (or `Infinity`) and `NaN` of
+the IEEE 754 non-finite values, optionally preceded by a `+` or `-` sign.
```
brandjon, line 2758: Nit: "The function also recognizes the names `Inf` (or `Infinity`) and `NaN`, optionally proceeded by a `+` or `-` sign. These construct the IEEE 754 non-finite values." (Avoids implication that there are only two non-finite values or that -inf and +inf share a name.)

spec.md
```diff
@@ -2606,6 +2742,24 @@ enumerate(["zero", "one", "two"])               # [(0, "zero"), (1, "one"), (2,
...
+Letter case is not significant.
+
+With no argument, `float()` returns `0.0`.
```
brandjon, line 2761: This is probably the correct behavior, but `int()` and `str()` are both errors (but `bool()` returns False). Fix by making those also return their zero type?

spec.md
```diff
@@ -2644,7 +2801,10 @@ implies `hash(x) == hash(y)`.
 `int(x[, base])` interprets its argument as an integer.
```
brandjon, line 2801: > a simple polynomial accumulator over the UTF-16 transcoding of the string

Ug. What will this do to compatibility with implementations not using UTF-16?

spec.md
```diff
@@ -2644,7 +2801,10 @@ implies `hash(x) == hash(y)`.
...
```
brandjon, line 2802: > `hash` fails if x, or any value upon which its hash depends, is unhashable.

If `hash` only works on strings, `x` can't depend on an unhashable value (or else it's already unhashable by virtue of its own type).

spec.md
```diff
@@ -2821,6 +2981,7 @@ All other strings, such as elements of a list of strings, are double-quoted.
 str(1)                          # '1'
 str("x")                        # 'x'
 str([1, "x"])                   # '[1, "x"]'
+str(0.0)                        # '0.0'        (formatted as if by "%g")
```
brandjon, line 2984: Mind the issue where Python's `%g` doesn't force the number to end in `.0` when equivalent to an integer. (If that's intentional, document it with `%g`.)

spec.md
```diff
@@ -2836,6 +2997,7 @@ type(x) returns a string describing the type of its operand.
 ```python
 type(None)              # "NoneType"
 type(0)                 # "int"
+type(0.0)               # "float"
 ```
 
 ### zip
```
brandjon, line 3003: The section on Grammar reference at the bottom needs updating. Likewise for tokens.

spec.md
```diff
@@ -273,27 +274,38 @@ x       index   starts_with     arg0
 ```
 
 *Literals*: literals are tokens that denote specific values.  Starlark
```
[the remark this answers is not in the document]
alandonovan: No, they are not literals. A literal is where the value is already in the syntax tree. By contrast, a list expression `[f(), g()]` is a computation. Its subexpressions may fail, and in the case of dict expressions, the insertion may fail due to duplicates (which is weird).

spec.md
```diff
@@ -273,27 +274,38 @@ x       index   starts_with     arg0
...
-has string and integer literals.
+has string, integer, and floating-point literals.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -273,27 +274,38 @@ x       index   starts_with     arg0
...
 
 ```text
 0                               # int
 123                             # decimal int
 0x7f                            # hexadecimal int
 0o755                           # octal int
 
+0.0     0.       .0             # float
+1e10    1e+10    1e-10
+1.1e10  1.1e+10  1.1e-10
```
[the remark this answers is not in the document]
alandonovan: Yes, but that's never useful. The grammar is unambiguous.

spec.md
```diff
@@ -493,6 +506,8 @@ remainder of floored division, respectively.
 If the signs of the operands differ, the sign of the remainder `x % y`
 matches that of the dividend, `x`.
 For all finite x and y (y ≠ 0), `(x // y) * y + (x % y) == x`.
+The `/` operator implements real division, and
```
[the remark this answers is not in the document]
alandonovan: Changed to "floating-point" since that's more accurate.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
 
 
+### Floating-point numbers
+
+The Starlark floating-point data type represents an IEEE 754
+double-precision floating-point number.
+Its [type](#type) is `"float"`.
+
+Arithmetic on floats using the `+`, `-`, `*`, `/`, `//`, and `%`
+operators follows the IEEE 754 standard.
```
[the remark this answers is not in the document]
alandonovan: I don't think we should get into this level of detail. Ultimately you will see your CPU's implementation of IEEE 754 with 64-bit rounding. That's good enough for Java, Go, Python and (most) C++ programmers.

spec.md
alandonovan: I have a separate TODO item to deal with that.

spec.md
alandonovan: It does this:
https://github.com/google/starlark-go/blob/master/starlark/library.go#L408

spec.md
alandonovan: Thanks, that was a relic of a more permissive hash function.

spec.md
alandonovan: Noted

The compact form `%g` is also used by `str(float)`. Its result                                       
uses the least precision required to accurately represent the value,                                 
omits unnecessary trailing zeros, and always contains a period                                       
and/or an exponent and thus unambiguously denotes a `float`, not an `int`.

spec.md
alandonovan: Done.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+However, computing the division or remainder of division by zero is a dynamic error.
```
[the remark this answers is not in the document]
alandonovan: Yes, but it depends on FPU state, and it differs from most major languages, so clearest to flag it as a difference.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+
+An arithmetic operation applied to a mixture of `float` and `int`
```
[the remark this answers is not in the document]
alandonovan: I think the meaning is plain enough from the previous paragraph. Comparisons are explained at length below, including the int<float case.

laurentlb:
A few thoughts on this

* nan == nan

That's interesting, because it removes the problems that appear when an object is not equal to itself (e.g. when used in a set or as a dict key).

* float < float is a transitive relation

This is nice, as it removes lots of weird cases.

* nan > inf

That seems to be a natural consequence of the previous choices, but this remains surprising. IEEE operations return nan when a computation is dubious. It seems to me that the alternative would be to raise an error (in IEEE 754, it seems that they use the nan value each time they would like to throw an exception, but they can't). In my experience, it's rare to have an intentional nan inside a computation. It's most likely a mistake, that should ideally be reported earlier.

Since nan is "not a number", maybe this value shouldn't support operators that expect numbers.

* implicit conversion from int to float

Lots of languages do that, but it's worth discussing it. Languages like C typically convert types to the "bigger" type. In Starlark, int are technically larger than floats. Of course, an implicit conversion to int would be even more surprising. Have you considered raising an error here and requiring explicit conversions?

In the past, we've often preferred raising errors in cases where no design was obvious. This allowed us to change the behavior later, once we have more data and user feedback (without breaking compatibility).

brandjon:
If you make it illegal to do regular arithmetic operations on NaN, you're really just deferring an error that should've been raised at the time the NaN was generated. I think the choice should be to either allow NaN arithmetic (including ordering), or eliminate NaN entirely.

The existence of NaN indicates a design philosophy that errors should not be show stoppers. It makes sense when you consider large matrix operations -- do you really want to guard every pixel's transformation with if statements, or replace bad pixels with dummy values? (Of course, that line of reasoning says that all operations should be defined, including division by zero, for which Python raises an exception.)

Ultimately, I think our motivation for adding floats to Starlark is configuration (e.g. possibly in rule attributes) and data interchange (protos, json), not numerical computation. This is why it makes sense to me to treat NaN as a singleton value (with `nan == nan`), rather than as a placeholder for "don't trust this computation" (with `nan != nan`). You want to test if the item you serialize to a file is NaN, not whether the computation of `a/b == c - d` where a and b are 0 and c and d are infinity. Likewise, you probably want to sort a bunch of values for dumping to a file, not necessarily to numerical walk them in a mathematically meaningful way.

The numerical use case could be handled by libraries and not in a first-class way. And you'd need an application-defined custom float type if you wanted to support overflowing to infinity without exception. And even then you still couldn't implement NaN unorderability because we want to require that even application-defined types use strict-weak orders (if they have ordering at all).

spec.md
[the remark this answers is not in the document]
brandjon: Discussed offline. Ultimately, the goal is not to fully determine the behavior across all implementations, but to simply specify that we use IEEE 754 representation and standard implementations of the basic arithmetic operators. We're not calling out details about rounding or what IEEE says about optional exceptions.

It is conceivable that the same floating point calculations on different Starlark implementations could produce (slightly) different results, and that's ok.

spec.md
brandjon: > omits unnecessary trailing zeros, and always contains a period and/or an exponent

Nit: This implies that it prints `5.` instead of `5.0`.

spec.md
brandjon: Hm. Seems like the Java interpreter is really throwing its weight around. Compare with what we did for strings, where neither implementation is favored over the other.

Why is it necessary that the hash of strings be fully specified across implementations, when their length and the bitsize of their elements is not?

brandjon:
Well now I really don't understand github's commenting model. You force-pushed, which usually destroys comment history by disassociating it from the code, but it appears to still be there. Maybe it does some diff cleverness. In any case, I suggest being wary of force-pushing again with many unresolved comments.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+operands works as if the `int` operand were first converted to a
+`float`.  For example, `3.141 + 1` is equivalent to `3.141 +
+float(1)`. The conversion may yield positive or negative
+infinity if the `int` value is too large to be represented as a `float`.
```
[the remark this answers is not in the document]
brandjon: This still needs updating: All attempts to create an overlarge float value result in overflow instead of inf.

spec.md
```diff
@@ -1827,8 +1960,20 @@ operand types are valid and how to convert the operand `x` to a string:
 s       any             as if by str(x)
 r       any             as if by repr(x)
 d       number          signed integer decimal
+o       number          signed octal
+e       number          float exponential format, lowercase (1.230000e+12)
+E       number          float exponential format, uppercase (1.230000E+12)
+f       number          float decimal format                (1230000000000.000000)
+F       number          same as %f
+g       number          compact format, lowercase           (0.0, 1.1, 1200, 1e+45, 1.2e+12) 
+G       number          compact format, lowercase           (0.0, 1.1, 1200, 1e+45, 1.2E+12)
```
brandjon, line 1969: Typo: uppercase

spec.md
```diff
@@ -1827,8 +1960,20 @@ operand types are valid and how to convert the operand `x` to a string:
...
 ```
 
+The compact form `%g` is also used by `str(float)`. Its result
+uses the least precision required to accurately represent the value,
+omits unnecessary trailing zeros in the mantissa, and always contains a period
```
brandjon, line 1974: Nit: omitting unnecessary trailing zeros means that `"%g" % 5.0 == "5."`, rather than `"5.0"`.

spec.md
brandjon: Resolved that these cases all fail.

spec.md
brandjon: (Forgot about this comment before I opened a new one, ah well.)

brandjon:
I made another pass to refresh my memory and close resolved comment threads. Let me know when I should look again -- and watch out for force pushes that may delink comments.

spec.md
alandonovan: In the course of building Herb, I encountered .bzl files whose effects were exquisitely dependent on the numbers returned by explicit `hash(str)` calls. Using an equally valid but different hash function changed execution so much that it exposed latent bugs in those files, or created significantly different build targets. This was my reluctant solution to the problem.

spec.md
alandonovan: Fixed.

spec.md
alandonovan: Now: "omits unnecessary trailing zeros in the significand"

alandonovan:
> > either allow NaN arithmetic (including ordering), or eliminate NaN entirely.
> 
> There's a middle ground where you can explicitly create the nan value and pass it to functions (e.g. application-specific functions, storing the value in a proto...), but it's not part of arithmetic operations (neither an input nor an output).

The point of NaN is much like the point of U+FFFD in Unicode: it provides a predictable way to deal with the result of a bad operation on one element, or a single screwed up value, that is an inevitable possibility of the data type's representation. This approach makes tools more reliable, especially when strings or numbers coming from outside the program have occasional errors.

spec.md
brandjon: That doesn't explain why hashes need to be specified across implementations of the interpreter. We don't need Bazel .bzl files to be processed by the Go implementation of Starlark. And even if we do, we can leave it out of the spec so that we're a little freer to change it later (at risk of practical incompatibilities from inertia).

spec.md
brandjon: Hair splitting, but that sounds the same as "uses the least precision required", and also doesn't seem to exclude `"5."`. You could just drop this middle clause, and expand the "contains a period" case to say a period with at least one following digit.

spec.md
alandonovan: On the contrary, it's a compelling reason why to specify hashes: unless the hash functions match exactly, Herb cannot simulate Blaze, and there is no workaround.

spec.md
alandonovan: Changed to:
```                                                                                             
Its result uses the least precision required to unambiguously represent the value,                              
omits unnecessary trailing zeros in the significand,                                                 
and the decimal point itself if the significand has no fraction,                                     
and always contains a decimal point or an exponent and thus                                          
unambiguously denotes a `float`, not an `int`.                                                       
```

alandonovan:
@brandjon, could you approve?

spec.md
alandonovan: This is not true: float x float arithmetic always yields a float, not an error (except division by zero, as mentioned).

spec.md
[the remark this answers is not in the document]
alandonovan: Done: "The implicit conversion fails if the `int` value is too large to be represented as a `float`."

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+`float`.  For example, `3.141 + 1` is equivalent to `3.141 +
+float(1)`. The conversion may yield positive or negative
+infinity if the `int` value is too large to be represented as a `float`.
+
+There are two floating-point division operators:
+`x / y ` yields the floating-point quotient of `x` and `y`,
+whereas `x // y` yields `floor(x / y)`, that is, the largest
+integer value not greater than `x / y`.
+Although the resulting number is integral, it is represented as a
+`float` if either operand is a `float`.
```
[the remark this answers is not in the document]
alandonovan: Done ("...that is, the largest representable integer value not greater than...")

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+
```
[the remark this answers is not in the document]
alandonovan: Done: The `%` operation computes the remainder of floored division.                                        
As with the corresponding operation on integers,                                                     
if the signs of the operands differ, the sign of the remainder `x % y`                               
matches that of the dividend, `x`.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+All float values are ordered, so they may be compared
+using operators such as `==` and `<`, and sorted using `sorted`.
+
+IEEE 754 defines two zero values, +0.0 and -0.0.
+They compare equal to each other.
+
+IEEE 754 defines two infinite float values `+Inf` and `-Inf`,
+which represent numbers greater/less than all finite float values.
+
+IEEE 754 defines many "not a number" (NaN) values.
```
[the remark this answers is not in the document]
alandonovan: No, Starlark has many NaN values: every 754 NaN value is a Starlark NaN value. If an application or library defines a floatbits function, you can observe them all. They just happen to compare equal in Starlark.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+They are non-finite, and represent the results of dubious operations
+such as `Inf/Inf`. All NaN values compare equal to each other,
+but greater than any non-NaN `float` value.
+(Starlark does not follow the IEEE 754 standard for NaN comparisons,
+which require that all comparisons with NaN are false, except NaN != NaN.)
+<!--
+This choice greatly simplifies the logic for float arithmetic by
```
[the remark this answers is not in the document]
alandonovan: That's not a spec issue, that's a Java API issue. If you want a type to be sortable, it must implement Comparable, and if it implements Comparable, it must be a strict weak order.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+ensuring many standard identities and invariants such as:
+- float < float is a transitive relation (a strict weak order)
```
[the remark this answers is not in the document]
alandonovan: Changed to:
```
- float < float and its inverse (<=) are transitive relations                                        
- float < float is a strict weak order: the relation eq is transitive,                               
  where eq(x, y) = not (x < y) and not (y < x).                                                      
```

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+- !(float < float) <=> (float >= float)
+- sorting a list of values that includes NaN is stable.
+
+Furthermore, implementations may assume that identical objects
+are equal, a useful optimization. Python in many cases exploits
+this optimization without the necessary invariant, leading to
+inconsistencies such as this:
+
+>>> nan == nan
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+
+A comparison operation may be applied to a mixture of int and float values.
+The result of such comparisons is mathematically exact, even if neither operand
+can be exactly represented by the type of the other.
+
+```python
+(type(1.0), type(1))            # ("float", "int")
+1.0 == 1			# True
+
+big = (1<<53)+1			# int not exactly representable as float
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+(big + 0.0) == big		# False (addition caused rounding down)
+(big + 0.0) - big		# 0.0   (both operands subject to rounding down)
+```
+
+Any bool, number, or string may be interpreted as a floating-point
+number by using the `float` built-in function.
+
+A float used in a Boolean context is considered true if it is
+non-zero.
```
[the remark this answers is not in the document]
alandonovan: Changed to:

A float used in a Boolean context is considered true if it is                                        
non-zero (not equal to 0.0 or -0.0). A NaN value is thus considered true.

spec.md
```diff
@@ -1170,8 +1279,8 @@ For example, an assignment statement updates the value held by a
 variable, and calls to some built-in functions such as `print` change
```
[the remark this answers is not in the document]
alandonovan: Removed.

spec.md
```diff
@@ -1170,8 +1279,8 @@ For example, an assignment statement updates the value held by a
...
 the state of the application that embeds the interpreter.
```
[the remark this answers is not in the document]
alandonovan: Changed to "Starlark has over a dozen core" for future-proofing. (It's true if you consider internal types like string iterators.)

spec.md
```diff
@@ -1429,11 +1538,11 @@ Lookup of locals and globals may fail if not yet defined.
 Starlark supports string literals of three different kinds:
 
 ```text
-Primary = int | string
+Primary = int | float | string
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -1429,11 +1538,11 @@ Lookup of locals and globals may fail if not yet defined.
...
 ```
 
-Evaluation of a literal yields a value of the given type (string, int) with the
-given value.
+Evaluation of a literal yields a value of the given type (string, int,
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -1543,14 +1652,15 @@ UnaryExpr = '+' Test
...
+- number        unary negation          (int, float)
+~ number        unary bitwise inversion (int)
 not x           logical negation        (any type)
 ```
 
-The `-` operator returns the opposite of any number.
-The `+` operator is never necessary in a correct program, but may
+The `+` and `-` operators may be applied to any number:
+`+` yields the operand unchanged, and `-` yields its negation.
+The `+` operator is never necessary in a correct program but may
```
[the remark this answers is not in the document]
alandonovan: Conceivably, if it's the only check standing between a dynamic error and launching the missiles.

spec.md
```diff
@@ -1666,18 +1776,29 @@ operator is its negation.
...
 tuple           # lexicographical
 list            # lexicographical
 ```
 
+Comparison of floating-point values follows the IEEE 754 standard
+for finite values (including -0.0) and for positive and negative
+infinity, but not for `NaN` values for which the standard behavior
+would break several mathematical identities. Thus:
+```
+-Inf < -1e50 < -1.0 < -1e-50 < 0 < 1e-50 < 1.0 < 1e50 < +Inf < NaN
```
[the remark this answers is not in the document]
alandonovan: It should be 0.0. Fixed.

spec.md
```diff
@@ -1666,18 +1776,29 @@ operator is its negation.
...
++0.0 == -0.0
+NaN == NaN
+```
+
 Applications may define additional types that support ordered
 comparison.
```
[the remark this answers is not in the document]
alandonovan: That's correct, but it needn't be said here. It's an issue of the Java or Go API.

spec.md
alandonovan: Not necessary. The "int or float:" comment above is to clarify patterns such as `number + number`, but here the patterns all say int.

spec.md
alandonovan: Good point. Added:

```
True and False may be converted to the values 0 and 1 using the `int` function,                      
but Booleans are not numbers.                                                                        
```

to Booleans.

spec.md
alandonovan: Done.

spec.md
alandonovan: Better. Done.

alandonovan:
I've addressed all the comments I overlooked before. (I still haven't got the hang of GH's code review system.) 

Great comments, all of them; thanks. PTAL.

spec.md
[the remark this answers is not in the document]
brandjon: Oh that's interesting. Perfectly fine as long as they compare/hash equal, but strange nonetheless.

spec.md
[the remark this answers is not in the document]
brandjon: Nit: `<=` isn't the inverse of `<`.

spec.md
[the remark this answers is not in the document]
brandjon: If Starlark the language permits non-strict-weak orders, then all the questions we raised about the semantics of sorted/min/max come back into play. Our choices are to either 1) ban non-strict-weak ordered types across all Starlark dialects, or 2) allow applications to have them but somehow underspecify the behavior of min/max/sorted over such values. In the latter case, we have to decide exactly how to underspecify them today as part of this change, even if there's no concrete plans for anyone to have such a type in the future.

spec.md
brandjon: It can't do it anyway if the string implementation is different, can it?

brandjon:
Left a few more replies. I think the only substantial issue remaining is whether we are specifying SWO as a requirement of all Starlark values. I think that it's reasonable to require this because it makes min/max/sorted make sense. Otherwise we either have to do a lot of work to produce exact specs for those (which we wanted to avoid, hence the tradeoffs we made for NaN in this spec); or else we have to be careful about exactly what the spec of those functions is in the presence of applicatoin-defined non-SWO types.

alandonovan:
> whether we are specifying SWO as a requirement of all Starlark values.

The spec states the behavior of the < and sorted operations on core data types, and grants the possibility of application-defined types having a defined order. The implementation defines the interface for application-defined types, and both implementations require strict weak orders---and given what we learned, that's really the only possible choice.

I don't think the spec needs to say anything more.

spec.md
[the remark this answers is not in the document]
alandonovan: Alright, I've added this:
```                                                                                        
Applications may define additional types that support ordered comparison.                            
The application-defined comparison relation must be a                                                
[strict weak ordering](https://en.wikipedia.org/wiki/Weak_ordering#Strict_weak_orderings).           
<!-- This is a prerequisite of the 'sorted' function. -->                                              
```

alandonovan (commented):
> I think the only substantial issue remaining is whether we are specifying SWO as a requirement of all Starlark values. I think that it's reasonable...

Alright, I've added the SWO requirement, as it must be true of all implementations.

alandonovan:
I saw only one addition comment (about ordering). Were there others?

alandonovan:
PTAL

spec.md
```diff
@@ -514,6 +529,100 @@ int("0xffff", 16)               # 65535
...
+
+IEEE 754 defines two infinite float values `+Inf` and `-Inf`,
+which represent numbers greater/less than all finite float values.
+
+IEEE 754 defines many "not a number" (NaN) values.
+They are non-finite, and represent the results of dubious operations
+such as `Inf/Inf`. All NaN values compare equal to each other,
+but greater than any non-NaN `float` value.
+(Starlark does not follow the IEEE 754 standard for NaN comparisons,
+which require that all comparisons with NaN are false, except NaN != NaN.)
```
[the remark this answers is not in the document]
brandjon: (Resolving as unnecessary)

spec.md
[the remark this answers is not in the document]
brandjon: (Moving comment to main thread)

brandjon:
Approved, thanks for bearing with me.

Nit on line 584:

> float < float and its inverse (<=) are transitive relations

Did you mean `>=`?

brandjon (approved):

alandonovan:
> Approved, thanks for bearing with me.
> 
> Nit on line 584:
> 
> > float < float and its inverse (<=) are transitive relations
> 
> Did you mean `>=`?

Updated to mention all five transitive comparison operators.

PR #125, opened 2020-11-11
starlark spec: document fail

### LINKED ISSUE #47: Add a built-in function similar to fail/assert
**Related:** https://github.com/google/starlark-go/pull/208#pullrequestreview-241974184

The Go implementation currently lacks a function similar to [fail()](https://docs.bazel.build/versions/master/skylark/lib/globals.html#fail) which is implemented in both Starlark Java and Rust. fail() is also used in both their test runners ([java](https://github.com/bazelbuild/bazel/blob/master/src/test/starlark/starlark_test.py#L81), [rust](https://github.com/google/starlark-rust/blob/master/starlark/tests/testutil/mod.rs#L87)) as they define assertion functions in every test and use it to throw errors when those assertions fail.

Go on the other hand uses a built-in [assert module](https://github.com/google/starlark-go/blob/master/starlarktest/assert.star) for test cases which include assert.eq(), assert.fails() and similar.
I think that a function similar to fail, or a more general one such as assert function/keyword would be useful to remove these inconsistencies. A function similar to assert for example could be used in the test cases of all 3 implementations and do the job of fail().

alandonovan:
Fixes #47

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2846,6 +2846,31 @@ a finite `float`.
 
 With no argument, `float()` returns `0.0`.
 
+### fail
+
+The `fail(*args)` function causes execution to fail
+with an error message that includes the string forms of the argument values.
+The precise formatting depends on the implementation.
+
+```python
+fail("oops")			# "fail: oops"
+fail("oops", 1, False)		# "fail: oops 1 False"
+```
+<!-- 
+Note: 
+
+Neither the template of the error message nor the formatting of the 
+values is prescribed here. Implementations may use a richer representation
+than str or repr, with additional debugging information.
+The error message is not observable by Starlark programs.
+
+The Java implementation also accepts two (deprecated) named parameters:
+- msg=value, an optional leading argument.
+- attr=value..., which adds an attribute prefix.
+and the Go implementation accepts a sep=... parameter, like print().
+See https://github.com/bazelbuild/starlark/issues/47.
+-->
+
 ### getattr
 
 `getattr(x, name[, default])` returns the value of the attribute (field or method) of x named `name`


laurentlb:
In Bazel, there's no `sep` argument. I'm not convinced we should mandate a formatting.

We could restrict fail to be `fail(obj)` and Starlark would simply return the object to the host application. The host application is free to do what it wants with the object.

alandonovan:
> In Bazel, there's no sep argument. I'm not convinced we should mandate a formatting.

I tend to agree (arguing against my words in https://github.com/bazelbuild/starlark/issues/47#issuecomment-496508485), but I was persuaded to keep it by your later comment "Using the same as print seems useful and consistent."

Let's remove sep then.

> We could restrict fail to be fail(obj) and Starlark would simply return the object to the host application. The host application is free to do what it wants with the object.

I would rather not do that because it creates the  expectation that the failure value is accessible to the catcher, and that's a hard invariant to maintain, in both implementations.

alandonovan:
PTAL

laurentlb:
As Jon mentioned (https://github.com/bazelbuild/starlark/issues/47#issuecomment-496615764), `print(x)` is not the same as `print(str(x))` in the Java implementation. It shows additional information. It may be useful for `fail` to do the same.

> the expectation that the failure value is accessible to the catcher, and that's a hard invariant to maintain

What's the problem here?

alandonovan:
> As Jon mentioned (#47 (comment)), print(x) is not the same as print(str(x)) in the Java implementation. It shows additional information. It may be useful for fail to do the same.

Yes, the current draft doesn't prescribe the exact formatting. It says "string forms" but not `str(x)`.

> > the expectation that the failure value is accessible to the catcher, and that's a hard invariant to maintain
> What's the problem here?

I was concerned that it requires errors never be flattened into their string forms, but I suppose this is already a requirement of correct reporting of the stack (which we now do). (Flattening into strings happens a lot but only for "leaf" exceptions in built-in functions, which have not yet been seen by the interpreter.) Nonetheless I would prefer not to mandate this feature, given that no-one has actually asked for it.

alandonovan:
PTAL

alandonovan:
@brandjon ping

spec.md
```diff
@@ -2606,6 +2606,24 @@ enumerate(["zero", "one", "two"])               # [(0, "zero"), (1, "one"), (2,
 enumerate(["one", "two"], 1)                    # [(1, "one"), (2, "two")]
 ```
 
+### fail
+
+The `fail(*args)` function causes execution to fail
+with an error message that includes the string forms of the argument values.
```
brandjon, line 2612: Explicitly mention that the form is not necessarily the same as what is returned by `str` or `repr`. (Optionally also explain that this is an intentional design choice to allow for information to be emitted by fail without necessarily exposing it to programmatic access by the Starlark code, e.g. non-deterministic information in deterministic applications like Bazel.)

brandjon (approved):

spec.md
alandonovan: In the interests of reducing non-normative text in the spec, I added this:

"""
...
The precise formatting depends on the implementation.                                                


"""

brandjon: Perfect.

PR #126, opened 2020-11-11
spec: permit assignment to empty sequence

### LINKED ISSUE #93: spec: allow assignment to empty tuple/list
Consider an assignment whose left-hand side is an empty tuple () or list [].

Python2 statically disallows () but allows [].
Python3 allows both.
Starlark in Go statically disallows both.
Starlark in Java dynamically disallows both.

What should the spec say?

I argue that the Python3 behavior is correct. It is simple, obvious, and consistent. It is even potentially useful. For example, () = x is a quick way to assert that x is an empty sequence. Another example: a code generator that emits a statement (v1, ..., vn) = x would need extra logic for the n = 0 case.

I propose that we remove the special treatment of an empty list/tuple.

@brandjon @laurentlb

alandonovan:
Fixes #93

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2162,12 +2162,12 @@ An assignment to a compound target checks that the right-hand value is a
 sequence with the same number of elements as the target.
 Each element of the sequence is then assigned to the corresponding
 element of the target, recursively applying the same logic.
-It is a static error if the sequence is empty.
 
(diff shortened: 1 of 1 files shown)


laurentlb (approved):
lgtm

PR #131, opened 2020-11-11
spec: print does not accept **kwargs

### LINKED ISSUE #22: spec: should print accept keyword arguments?
The spec https://github.com/bazelbuild/starlark/blob/master/spec.md#print says print accepts keyword arguments. This text came from the Go implementation, whose behavior it describes. The Java implementation rejects keyword arguments to print.  We should decide whether we prefer the spec as is, or the Java implementation, and make all docs and code consistent with that. Neither version of Python allows kwargs, so probably the right thing do to is disallow it.

alandonovan:
Fixes #22

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2707,7 +2707,7 @@ min("two", "three", "four", key=len)            # "two", the shortest
 
 ### print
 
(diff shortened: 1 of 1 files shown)


laurentlb (approved):
lgtm

PR #132, opened 2020-11-11
starlark: integers are exact

### LINKED ISSUE #120: Arithmetic seems to be under-specified
ints and arithmetic on them seem to be a little bit loosy-goosy in the spec

- The size of int is implementation defined, but there is no minimum required size. Hypothetically, `3 + 3 = <dynamic error>` would be compliant. A reasonable minimum might be 32 or 64 bits.
- There are ways to manipulate the bit-representation of an int, but the spec doesn't say what that is. The spec should likely define int as two's complement. (Yeah I know this is kinda nitpicky, since two's complement is practically universal)
- It's not specified whether shifts are arithmetic (i.e. sign-extending) or logical. Python shifts are purely arithmetic (`x << n` is a shorthand for `x * 2**n`, `x >> n` is short for `x // 2**n`).
- Shift amounts larger than the size of int should probably be a dynamic error (currently: "Implementations may impose a limit on the second operand of a left shift.")
- Starlark does not have Python's `**` operator.

alandonovan:
Fixes #120

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -414,7 +414,8 @@ These are the main data types built in to the interpreter:
 ```text
 NoneType                     # the type of None
 bool                         # True or False
-int                          # a signed integer
+int                          # a signed integer of arbitrary magnitude
+float                        # an IEEE 754 double-precision floating-point number
 string                       # a byte string
 list                         # a fixed-length sequence of values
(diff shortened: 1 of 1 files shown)


laurentlb (approved):
lgtm

PR #133, opened 2020-11-13
spec: duplicate keyword args f(x=1, x=2) is a static error

### LINKED ISSUE #21: spec: statically reject repeated keyword arguments in call
I notice that both Python2 and Python3 statically reject repeated keyword arguments in a call:
```
% python2 -c 'if False: f(a=1, a=1)'
  File "<string>", line 1
SyntaxError: keyword argument repeated
% python3 -c 'if False: f(a=1, a=1)'
  File "<string>", line 1
SyntaxError: keyword argument repeated
```
Perhaps Starlark should also.  Neither the Go nor Java implementation currently does.

(Note: this issue is independent of the question of whether repeated arguments are checked dynamically when a **kwargs is present in the call.)

alandonovan:
This wording comes from https://github.com/google/starlark-go/commit/2c65f9e0f5e155edb22bcf767d709321069553e9

Fixes #21

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -942,6 +942,11 @@ Once the parameters have been successfully bound to the arguments
 supplied by the call, the sequence of statements that comprise the
 function body is executed.
 
+It is a static error if a function call has two named arguments of the
+same name, such as `f(x=1, x=2)`. A call that provides a `**kwargs`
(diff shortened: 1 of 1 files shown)


laurentlb (approved):
lgtm

PR #152, opened 2021-01-19
spec: fix mistake regarding sign of x % y

alandonovan:
The Go and Java implementations agree with Python 2 and 3 that
the sign of x % y should match the divisor (y) not the dividend (x).
This change makes the spec match the implementations and the intent.


laurentlb (approved):
LGTM

PR #154, opened 2021-01-22
spec: document %x and %X conversions for string % int

alandonovan:
See cr/353243297 for Java implementation change. The Go implementation already supports them.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2040,7 +2040,9 @@ operand types are valid and how to convert the operand `x` to a string:
 s       any             as if by str(x)
 r       any             as if by repr(x)
 d       number          signed integer decimal
-o       number          signed octal
+o       number          signed octal, no 0o prefix
+x       number          signed hexadecimal, lowercase, no 0x prefix
+X       number          signed hexadecimal, uppercase, no 0x prefix
 e       number          float exponential format, lowercase (1.230000e+12)
 E       number          float exponential format, uppercase (1.230000E+12)
 f       number          float decimal format                (1230000000000.000000)


spec.md
```diff
@@ -2041,6 +2041,8 @@ s       any             as if by str(x)
 r       any             as if by repr(x)
 d       number          signed integer decimal
 o       number          signed octal
```
brandjon, line 2043: Optional: Clarify that no preceding `0x` or `0o` notation is emitted?

brandjon (approved):

brandjon:
> It is an error if the argument does not have the type required by the conversion specifier, except that ints may converted to floats and floats may truncated to ints.

This appears to differ from Python 3. Is that intentional? It truncates for `%d` but not the other bases.

```
>>> "%d" % 1.1
'1'
>>> "%x" % 1.1
TypeError: %x format: an integer is required, not float
>>> "%o" % 1.1
TypeError: %o format: an integer is required, not float
```

brandjon:
Note that Python's `hex()`, `oct()`, and `bin()` functions also do not accept floats.

alandonovan:
Yes, intentional. Python2 rounds floats when formatting as an integer. Python3 does so for decimal, but not for the other bases, which seems inconsistent to me.

spec.md
alandonovan: Good point; done.

The usual way to get the prefix is `'%#x'` but Starlark formatting doesn't yet support flags. (`'0x%x'` is wrong for negative numbers.)

brandjon:
I suppose the counterargument is that base conversion is almost always intended for integers. But I certainly don't feel strongly enough to choose between API consistency and Python-Starlark consistency, so I defer to you. :)

Issue #112, opened 2020-10-02
spec: new 'bytes' data type


alandonovan:
This proposal argues for the addition of a new 'bytes' data type to Starlark, defined as an immutable sequence of byte (uint8) values.

Starlark has implementations in Go, Java, and Rust. It is crucial for performance that Starlark use the same string data type as the host language, but this poses a problem, because Java strings are sequences of uint16 values, conventionally holding UTF-16 encoded text, whereas Go's strings are sequences of uint8 values, conventionally holding UTF-8 encoded text.

The solution to this problem is to specify the behavior of Starlark strings in terms of an implementation-defined parameter K, which is 8 for Go and 16 for Java. Nearly all string operations over text can be usefully described this way, enabling portable text handling. (The sole obvious exception is `str.strip(chars='abc')`, as it treats the chars string as a set rather than an ordered sequence.)   [This needs its own proposal. TODO(adonovan)]

This leaves the Java implementation without a way to represent arbitrary byte strings, such as the contents of arbitrary binary files, or fields of protocol buffers. This need would be met by adding a new data type, 'bytes', to the core language:

- the literal syntax `b"ABC"` would denote a byte string, analogous to how "ABC" denotes a UTF-K string.
- byte string literals would permit hex escapes `\xXX` and octal escapes `\377` over the range [0-255]. Ordinary strings would permit hex and octal escapes only for the ASCII range [0-127], but would additionally allow \uXXXX and \UXXXXXXXX escapes to denote the UTF-K encoding of a unicode code point specified in hex.
- `type(bytes)` would return `"bytes"`.
- `len(bytes)` would return the number of bytes.
- `bool(bytes)` would be equivalent to `len(bytes) > 0`.
- indexing `bytes[int]` would yield a 1-element byte string.
- the following string methods would have byte-string counterparts: [count endswith find index join lstrip partition replace rfind rindex rpartition rsplit rstrip split startswith strip]. This set excludes methods related to textual concepts such as letter vs number, or upper case vs lower. We should perhaps start with smaller set.
- the elems method would iterate over the 1-byte substrings, and elem_ords would iterate over the numeric byte values.
- `str(bytes)` would decode bytes to UTF-K, replacing invalid sequences by U+FFFD.
- 'bytes(str)' would encode a text string to UTF-8, replacing invalid code units with the encoding of U+FFFD.
- repr(bytes) would return a byte string literal such as `b"ABC"`.

brandjon:
> byte string literals

[...]

PR #161, opened 2021-02-05
spec: bytes data type

alandonovan:
This change adds initial specification of the bytes data type
following length discussion in #112.
It also explains the implementation-dependent encoding of
text strings, and the \u \U \X escapes.

More will follow, but let's get the easy parts out of the way first.

Updates #112

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -385,21 +385,14 @@ It is an error if the value of an octal or hexadecimal escape is greater than de
 "\x41-\x5A"             # "A-Z"
 ```
 
-The bit width of each string element is not defined by this spec,
-and for reasons of efficiency and interoperability with the host language,
-it varies across implementations.
-For example, in the Go and Rust implementations,
-each string element is an 8-bit value (a byte) and Unicode text is encoded as UTF-8,
-whereas in the Java implementation,
-string elements are 16-bit values (Java `char`s) and Unicode text is encoded as UTF-16.
-This spec generalizes both approaches by referring to the string encoding as "UTF-K",
-where the value of K is implementation-defined.
-
-A *Unicode escape* denotes the UTF-K encoding of a single Unicode code point.
+A *Unicode escape* denotes the UTF-K encoding of a single, valid Unicode code point,
+where K is the implementation-defined number of bits in each string element
+(see [strings](#strings)).
 The `\uXXXX` form, with exactly four hexadecimal digits,
 denotes a 16-bit code point, and the `\UXXXXXXXX`,
 with exactly eight digits, denotes a 32-bit code point.
-It is an error if the value lies in the surrogate range (U+D800 to U+DFFF).
+It is an error if the value lies in the surrogate range (U+D800 to U+DFFF)
+or is greater than U+10FFFF.
 
 ```python
 '\u0041'		# "A", an ASCII letter (U+0041)
@@ -409,7 +402,7 @@ It is an error if the value lies in the surrogate range (U+D800 to U+DFFF).
 ```
 
 The length of the encoding of a single Unicode code point may vary
-based on the implementation's string encoding ("K").
+based on the implementation's value of K:
 
 ```python
 len("A") 		# 1
@@ -418,19 +411,13 @@ len("界")               # 3 (UTF-8) or 1 (UTF-16)
 len("😀")               # 4 (UTF-8) or 2 (UTF-16)
 ```
 
-An *element escape* denotes a single element of a text string.
-It consists of `\X`, followed by two hexadecimal digits if K=8, or four if K=16.
-Element escapes are needed only to express strings that do not
-contain a valid encoding of text, such as the result of a substring
-operation that truncates an incomplete UTF-K encoding.
-Element escapes are inherently non-portable,
-and their use in source code is discouraged.
-
-```python
-"\X41"			# 'A', same as '\x41'
-"\XFF"			# '\XFF', not a valid text string
-"😀"[:1]		# "\Xf0" (UTF-8) or "\Xd83d" (UTF-16)
-```
+Although string values may be capable of representing any sequence elements,
+string  _literals_ can denote only sequences of UTF-K code
+units that are valid encodings of text.
+(Any literal syntax capable of representing arbitrary element sequences
+would inherently be non-portable across implementations.)
+Consequently, when the `repr` function is applied to a string
+containing an invalid encoding, its result is not a valid string literal.
 
 An ordinary string literal may not contain an unescaped newline,
 but a *multiline string literal* may spread over multiple source lines.
@@ -498,8 +485,6 @@ with the following differences:
 
 - Octal and hexadecimal escapes may specify any byte value from
   zero (`\000` or `\x00`) to 255 (`\377` or `\xFF`).
-
-- Element escapes `\X` are not permitted.
 
 - A Unicode escape `\uXXXX` or `\UXXXXXXXX` denotes the byte
   sequence of the UTF-8 encoding of the specified 16- or 32-bit code point.
@@ -734,18 +719,24 @@ float(3) / 2                                    # 1.5
 A string is an immutable sequence of elements that encode Unicode text.
 The [type](#type) of a string is `"string"`.
 
-Depending on the implementation, the elements may be 8- or 16-bit values,
-and may hold arbitrary values of the element type;
-we call the implementation's element width K.
-By convention, however, strings are typically used to hold valid encodings
-of Unicode code points, encoded using UTF-K.
+For reasons of efficiency and interoperability with the host language,
+the number of bits in each string element, which we call K,
+is specified to be either 8 or 16, depending on the implementation.
+For example, in the Go and Rust implementations,
+each string element is an 8-bit value (a byte) and Unicode text is encoded as UTF-8,
+whereas in the Java implementation,
+string elements are 16-bit values (Java `char`s) and Unicode text is encoded as UTF-16.
+
+An implementation may permit strings to hold arbitrary values of the element type,
+including sequences that do not denote encode valid Unicode text;
+or, it may disallow invalid sequences, and operations that would form them.
 
 The built-in `len` function returns the number of elements in a string.
 
 Strings may be concatenated with the `+` operator.
 
 The substring expression `s[i:j]` returns the substring of `s` from
-index `i` up to index `j`.
+element index `i` up to index `j`.
 <!-- TODO: The Rust implementation of s[i:j] may fail if it cuts a 
      UTF-8 sequence in half. Need to accommodate that here. -->
 The index expression `s[i]` returns the
@@ -755,7 +746,8 @@ Strings are hashable, and thus may be used as keys in a dictionary.
 
 Strings are totally ordered lexicographically, so strings may be
 compared using operators such as `==` and `<`.
-(Beware that the UTF-16 string encoding is not order-preserving.)
+(Beware that the UTF-16 string encoding is not order-preserving
+with respect to code point values.)
 
 Strings are _not_ iterable sequences, so they cannot be used as the operand of
 a `for`-loop, list comprehension, or any other operation than requires
@@ -807,16 +799,16 @@ Strings have several built-in methods:
 A _bytes_ is an immutable sequence of values in the range 0-255.
 The [type](#type) of a bytes is `"bytes"`.
 
-Unlike a string, a bytes may represent binary data,
+Unlike a string, which is intended for text, a bytes may represent binary data,
 such as the contents of an arbitrary file, without loss.
 
 The built-in `len` function returns the number of elements (bytes) in a `bytes`.
 
 Two bytes values may be concatenated with the `+` operator.
 
-The slice expression `b[i:j]` returns the subsequence of `b` from
-index `i` up to index `j`.  The index expression `b[i]` returns the
-int value of the ith element.
+The slice expression `b[i:j]` returns the subsequence of `b`
+from index `i` up to but not including index `j`.
+The index expression `b[i]` returns the int value of the ith element.
 
 Like strings, bytes are hashable, totally ordered, and not iterable,
 and are considered True if they are non-empty. 
@@ -831,7 +823,7 @@ TODO(https://github.com/bazelbuild/starlark/issues/112)
     replace
 - specify ord, chr?
 - hash(bytes)
-- support 'bytes in bytes'?
+- support 'bytes in bytes', 'int in bytes'?
 - bytes(...) function
 - encode, decode methods?
 - can we reduce string iterator methods without loss of generality/efficiency?
@@ -960,7 +952,7 @@ Dictionaries provide constant-time operations to insert an element, to
 look up the value for a key, or to remove an element.  Dictionaries
 are implemented using hash tables, so keys must be hashable.  Hashable
 values include `None`, Booleans, numbers, strings, and bytes, and tuples
-composed from hashable values.  Most mutable values, such as lists,
+composed from hashable values.  Most mutable values, such as lists
 and dictionaries, are not hashable, unless they are frozen.
 Attempting to use a non-hashable value as a key in a dictionary
 results in a dynamic error.
@@ -1492,7 +1484,7 @@ variable, and calls to some built-in functions such as `print` change
 the state of the application that embeds the interpreter.
 
 Values of some data types, such as `NoneType`, `bool`, `int`, `float`,
-`string` and `bytes`, are _immutable_; they can never change.
+`string`, and `bytes`, are _immutable_; they can never change.
 Immutable values have no notion of _identity_: it is impossible for a
 Starlark program to tell whether two integers, for instance, are
 represented by the same object; it can tell only whether they are
@@ -1747,14 +1739,13 @@ Lookup of locals and globals may fail if not yet defined.
 
 ### Literals
 
-Starlark supports string literals of four different kinds:
+Starlark supports literals of four different kinds:
 
 ```text
 Operand = int | float | string | bytes
 ```
 
-Evaluation of a literal yields a value of the given type
-(int, float, string, or bytes) with the given value.
+Evaluation of an int, float, string, or bytes literal yields the value of that literal.
 See [Literals](#lexical elements) for details.
 
 ### Parenthesized expressions
@@ -2054,7 +2045,7 @@ Concatenation
      list + list
     tuple + tuple
 
-Repetition (bytes/string/list/tuple)
+Repetition (string/bytes/list/tuple)
       int * sequence
  sequence * int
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -702,6 +802,41 @@ Strings have several built-in methods:
 * [`upper`](#string·upper)
 
 
+### Bytes
+
+A _bytes_ is an immutable sequence of values in the range 0-255.
+The [type](#type) of a string is `"string"`.
```
illicitonion, line 808:
```suggestion
The [type](#type) of a bytes is `"bytes"`.
```

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
 
 ### Strings
 
-A string represents an immutable sequence of bytes.
+A string is an immutable sequence of elements that encode Unicode text.
 The [type](#type) of a string is `"string"`.
 
-Strings can represent arbitrary binary data, including zero bytes, but
-most strings contain text, encoded by convention using UTF-8.
+Depending on the implementation, the elements may be 8- or 16-bit values,
```
illicitonion, line 737: I'm not sure what the intended use of these strings is, if they may contain invalid data, and they index the same as `bytes`... What's the goal of having two distinct types here, where their guarantees and API surface are ~identical? Is it just to give the Java implementation scope to use UTF-16 to represent strings?

illicitonion: Concretely, I guess my question really is: Should strings perhaps guarantee they contain valid UTF-K data, and perhaps index or at least allow views to iterate over codepoints rather than k-byte elements? (Or is some of this the "still to come" part? :))

illicitonion:
I've deleted a question which I left on this PR, as reading through the linked issue gave me the needed context - apologies for not reading thoroughly enough the first time!

spec.md
alandonovan: Done.

spec.md
```diff
@@ -365,15 +368,68 @@ allowing a long string to be split across multiple lines of the source file.
 def"			# "abcdef"
 ```
 
-An *octal escape* encodes a single byte using its octal value.
+An *octal escape* encodes a single string element using its octal value.
 It consists of a backslash followed by one, two, or three octal digits [0-7].
```
brandjon, line 372: > followed by one, two, or three octal digits

This is precedent for the problem we discussed earlier today about `\XFF` vs `\XFFFF`. We could just consume four hex digits if possible, then three if possible, and finally two, erroring out if that's not possible. Then you don't need to know whether k=8 or k=16 to parse `\X` escapes, and you aren't forced to write `\X00FF` instead of `\XFF`.

Still doesn't change the other problems we discussed about the ambiguity of the denotation, I think, so I guess it's moot.

spec.md
```diff
@@ -365,15 +368,68 @@ allowing a long string to be split across multiple lines of the source file.
...
+whereas in the Java implementation,
+string elements are 16-bit values (Java `char`s) and Unicode text is encoded as UTF-16.
+This spec generalizes both approaches by referring to the string encoding as "UTF-K",
+where the value of K is implementation-defined.
+
+A *Unicode escape* denotes the UTF-K encoding of a single Unicode code point.
+The `\uXXXX` form, with exactly four hexadecimal digits,
+denotes a 16-bit code point, and the `\UXXXXXXXX`,
+with exactly eight digits, denotes a 32-bit code point.
+It is an error if the value lies in the surrogate range (U+D800 to U+DFFF).
```
brandjon, line 402: Also an error if it's past U+10FFFF.

spec.md
```diff
@@ -365,15 +368,68 @@ allowing a long string to be split across multiple lines of the source file.
...
+based on the implementation's string encoding ("K").
+
+```python
+len("A") 		# 1
+len("Д") 		# 2 (UTF-8) or 1 (UTF-16)
+len("界")               # 3 (UTF-8) or 1 (UTF-16)
+len("😀")               # 4 (UTF-8) or 2 (UTF-16)
+```
+
+An *element escape* denotes a single element of a text string.
```
brandjon, line 421: Comment out or delete this part as per yesterday's discussion.

brandjon (commented):
(Still reviewing)

spec.md
```diff
@@ -1358,7 +1492,7 @@ variable, and calls to some built-in functions such as `print` change
 the state of the application that embeds the interpreter.
 
 Values of some data types, such as `NoneType`, `bool`, `int`, `float`,
-and `string`, are _immutable_; they can never change.
+`string` and `bytes`, are _immutable_; they can never change.
```
brandjon, line 1495: Nit: oxford comma

spec.md
```diff
@@ -1613,14 +1747,14 @@ Lookup of locals and globals may fail if not yet defined.
 
 ### Literals
 
-Starlark supports string literals of three different kinds:
+Starlark supports string literals of four different kinds:
```
brandjon, line 1750: Delete "string" on this line?

spec.md
```diff
@@ -1613,14 +1747,14 @@ Lookup of locals and globals may fail if not yet defined.
...
 
 ```text
-Operand = int | float | string
+Operand = int | float | string | bytes
 ```
 
 Evaluation of a literal yields a value of the given type
-(int, float, or string) with the given value.
+(int, float, string, or bytes) with the given value.
```
brandjon, line 1757: "given value" -> "denoted content"?

spec.md
```diff
@@ -1915,10 +2050,11 @@ Bitwise operations:
 
 Concatenation
    string + string
+    bytes + bytes
      list + list
     tuple + tuple
 
-Repetition (string/list/tuple)
+Repetition (bytes/string/list/tuple)
```
brandjon, line 2057: Nit: string first, then bytes, to match order for concatenation (and elsewhere)?

spec.md
```diff
@@ -2250,7 +2386,7 @@ f("n")                                          # 2
 ### Index expressions
 
 An index expression `a[i]` yields the `i`th element of an _indexable_
```
brandjon, line 2388: "Element" needs some clarification here. A bytes element is a byte, but there is no datatype called a "byte". We can also note that `a[i]` is equal to `a[i:i+1]` for strings alone.

spec.md
```diff
@@ -2250,7 +2386,7 @@ f("n")                                          # 2
 ### Index expressions
```
brandjon, line 2386: Above: Update "Membership tests" to specify the `in` / `not in` operators for bytes. `65 in b"A"` returns True in Python.

spec.md
```diff
@@ -2292,7 +2428,7 @@ type, such as a tuple or string, or a frozen value of a mutable type.
 ### Slice expressions
```
brandjon, line 2428: Below:

> Slicing a tuple or string may be more efficient than slicing a list

Add bytes to that.

brandjon:
Driveby: "type(x) returns a string describing the type of its operand." -> backticks around `type(x)`.

brandjon:
Also:

```
Operand = identifier
        | int | float | string
        | ListExpr | ListComp
        | DictExpr | DictComp
        | '(' [Expression [',']] ')'
```

Needs a `| bytes` there.

spec.md
alandonovan: During discussions, we have decided to abandon `\X` element escapes, since they are inherently non-portable, so this is indeed moot.

spec.md
```diff
@@ -365,15 +368,68 @@ allowing a long string to be split across multiple lines of the source file.
...
+'\x00'			# "\x00"  a string containing a single NUL element
+'\0A'			# "\n"    hexadecimal A = decimal 10
+"\x41-\x5A"             # "A-Z"
+```
+
+The bit width of each string element is not defined by this spec,
+and for reasons of efficiency and interoperability with the host language,
+it varies across implementations.
+For example, in the Go and Rust implementations,
+each string element is an 8-bit value (a byte) and Unicode text is encoded as UTF-8,
```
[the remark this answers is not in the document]
alandonovan: Good point. Moved and smoothed.

spec.md
```diff
@@ -365,15 +368,68 @@ allowing a long string to be split across multiple lines of the source file.
...
+whereas in the Java implementation,
+string elements are 16-bit values (Java `char`s) and Unicode text is encoded as UTF-16.
+This spec generalizes both approaches by referring to the string encoding as "UTF-K",
+where the value of K is implementation-defined.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
alandonovan: Done.

spec.md
alandonovan: Done.

spec.md
```diff
@@ -419,6 +475,41 @@ b"		# "a\\\nb"
 It is an error for a backslash to appear within a string literal other
```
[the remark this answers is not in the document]
alandonovan: > Do we specify what constitutes a line ending in a Starlark source file? Specifically, what the algorithm is for converting a raw U+000D or U+000A or a pair of them, into a single U+000A?

What more needs to be said? The scanner needs to recognize line endings (however they are defined by the platform), in three places:
1) escaped, in which case are ignored;
2) unescaped, outside a string literal, where they make a NEWLINE token;
2) unescaped, in a multiline string literal, where they make a \n (as the quoted paragraph explains).

spec.md
```diff
@@ -419,6 +475,41 @@ b"		# "a\\\nb"
...
+A raw bytes literal may be indicated by either a `br` or `rb` prefix.
+
+Non-escaped text within a bytes literal denotes the UTF-8 encoding of that text.
+Bytes literals support the same escape sequences as text strings,
+with the following differences:
+
+- Octal and hexadecimal escapes may specify any byte value from
+  zero (`\000` or `\x00`) to 255 (`\377` or `\xFF`).
+
+- Element escapes `\X` are not permitted.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
```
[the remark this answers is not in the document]
alandonovan: I haven't implemented int(bytes), and I don't think we should, though it would be easy to do.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
...
 ### Strings
 
-A string represents an immutable sequence of bytes.
+A string is an immutable sequence of elements that encode Unicode text.
 The [type](#type) of a string is `"string"`.
 
-Strings can represent arbitrary binary data, including zero bytes, but
-most strings contain text, encoded by convention using UTF-8.
+Depending on the implementation, the elements may be 8- or 16-bit values,
+and may hold arbitrary values of the element type;
```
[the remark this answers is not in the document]
alandonovan: Let's leave Rust for a follow-up, though I have left wiggle room (or hidden TODO comments) in anticipation.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
...
+we call the implementation's element width K.
+By convention, however, strings are typically used to hold valid encodings
+of Unicode code points, encoded using UTF-K.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
...
 
-The built-in `len` function returns the number of bytes in a string.
+The built-in `len` function returns the number of elements in a string.
 
 Strings may be concatenated with the `+` operator.
 
 The substring expression `s[i:j]` returns the substring of `s` from
-index `i` up to index `j`.  The index expression `s[i]` returns the
-1-byte substring `s[i:i+1]`.
+index `i` up to index `j`.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
...
+<!-- TODO: The Rust implementation of s[i:j] may fail if it cuts a 
+     UTF-8 sequence in half. Need to accommodate that here. -->
```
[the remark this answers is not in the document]
alandonovan: I agree, it seems tricky. Let's leave this in a comment until the Rust folks have some experience and know what they want us to say here.

spec.md
```diff
@@ -639,28 +731,36 @@ float(3) / 2                                    # 1.5
...
+The index expression `s[i]` returns the
+1-element substring `s[i:i+1]`.
 
 Strings are hashable, and thus may be used as keys in a dictionary.
 
 Strings are totally ordered lexicographically, so strings may be
 compared using operators such as `==` and `<`.
+(Beware that the UTF-16 string encoding is not order-preserving.)
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -702,6 +802,41 @@ Strings have several built-in methods:
 * [`upper`](#string·upper)
 
 
+### Bytes
+
+A _bytes_ is an immutable sequence of values in the range 0-255.
+The [type](#type) of a bytes is `"bytes"`.
+
+Unlike a string, a bytes may represent binary data,
+such as the contents of an arbitrary file, without loss.
```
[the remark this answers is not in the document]
alandonovan: Changed to: `Unlike a string, which is intended for text, ...`, using fewer words so as not to obscure the main point in technical details that are explained elsewhere.

spec.md
```diff
@@ -702,6 +802,41 @@ Strings have several built-in methods:
...
+
+The built-in `len` function returns the number of elements (bytes) in a `bytes`.
+
+Two bytes values may be concatenated with the `+` operator.
+
+The slice expression `b[i:j]` returns the subsequence of `b` from
+index `i` up to index `j`.  The index expression `b[i]` returns the
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -824,7 +959,7 @@ The [type](#type) of a dictionary is `"dict"`.
 Dictionaries provide constant-time operations to insert an element, to
 look up the value for a key, or to remove an element.  Dictionaries
 are implemented using hash tables, so keys must be hashable.  Hashable
-values include `None`, Booleans, numbers, and strings, and tuples
+values include `None`, Booleans, numbers, strings, and bytes, and tuples
 composed from hashable values.  Most mutable values, such as lists,
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
alandonovan: Done.

spec.md
alandonovan: Good catch!

spec.md
alandonovan: "Evaluation of an int, float, string, or bytes literal yields the value of that literal."

spec.md
alandonovan: Done.

spec.md
alandonovan: It's in the phase 2 TODO list under bytes.

spec.md
alandonovan: The bytes type is defined as:

> A _bytes_ is an immutable sequence of values in the range 0-255.

and shortly after, it says:

> The index expression `b[i]` returns the int value of the ith element.

spec.md
alandonovan: Done.

alandonovan (commented):
Thanks Jon; PTAL.

spec.md
```diff
@@ -385,21 +385,12 @@ It is an error if the value of an octal or hexadecimal escape is greater than de
...
 
-The bit width of each string element is not defined by this spec,
-and for reasons of efficiency and interoperability with the host language,
-it varies across implementations.
-For example, in the Go and Rust implementations,
-each string element is an 8-bit value (a byte) and Unicode text is encoded as UTF-8,
-whereas in the Java implementation,
-string elements are 16-bit values (Java `char`s) and Unicode text is encoded as UTF-16.
-This spec generalizes both approaches by referring to the string encoding as "UTF-K",
-where the value of K is implementation-defined.
```
brandjon: We can't push the definition of "UTF-K" below its first use.

spec.md
```diff
@@ -734,18 +717,24 @@ float(3) / 2                                    # 1.5
 A string is an immutable sequence of elements that encode Unicode text.
 The [type](#type) of a string is `"string"`.
 
-Depending on the implementation, the elements may be 8- or 16-bit values,
-and may hold arbitrary values of the element type;
-we call the implementation's element width K.
-By convention, however, strings are typically used to hold valid encodings
-of Unicode code points, encoded using UTF-K.
+The number of bits in each string element, which we call K, is not defined by this spec,
```
brandjon, line 720: "is not defined by this spec" --> "is specified to be either 8 or 16, depending on the implementation". Avoid the implication that 8 or 16 is a suggestion and not a mandate.

spec.md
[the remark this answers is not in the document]
brandjon: The scanner is part of the spec, no? I believe Python defines newlines in a platform-independent way, and we should probably do the same. A raw `\r\n` on unix should still produce a single `\n`, not a `\r\n`, inside a multiline string literal.

spec.md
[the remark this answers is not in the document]
brandjon: Ah, right, that's the string interpretation of ascii numerals. Nevermind, I agree we shouldn't do it.

spec.md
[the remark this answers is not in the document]
brandjon: A TODO is fine for unblocking this PR.

brandjon (approved):

spec.md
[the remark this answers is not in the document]
alandonovan: My point is that the only place that needs to take a stance on the concrete representation of a line ending is case 3, which already spells it out thus:
                                                                                                                      
> Regardless of the platform's convention for text line endings---for                                                           
> example, a linefeed (\n) on UNIX, or a carriage return followed by a                                                          
> linefeed (\r\n) on Microsoft Windows---an unescaped line ending in a                                                          
> multiline string literal always denotes a line feed (\n).

spec.md
alandonovan: Now:
"""
For reasons of efficiency and interoperability with the host language,                                                        
the number of bits in each string element, which we call K,                                                                   
is specified to be either 8 or 16, depending on the implementation.
"""

alandonovan:
> Driveby: "type(x) returns a string describing the type of its operand." -> backticks around `type(x)`.

Done.

> Needs a `| bytes` there.

Done.

> The scanner is part of the spec, no? I believe Python defines newlines in a platform-independent way, and we should probably do the same. A raw \r\n on unix should still produce a single \n, not a \r\n, inside a multiline string literal.
                      
Yes. We say that here:
"""                                                                                   
Regardless of the platform's convention for text line endings---for                                                           
example, a linefeed (\n) on UNIX, or a carriage return followed by a                                                          
linefeed (\r\n) on Microsoft Windows---an unescaped line ending in a                                                          
multiline string literal always denotes a line feed (\n).                                                                     
"""

spec.md
alandonovan: Good point: I forgot that the literals are presented before the data type they represent.

Tweaked.

PR #164, opened 2021-02-12
spec: more bytes operations

alandonovan:
spec: more bytes operations
    
    This change defines the semantics of: 
    - str(bytes) - UTF-8 decoding with U+FFFD replacement
    - bytes(str) - UTF-8 encoding with U+FFFD replacement
    - bytes.elems() -- iterable of int values of byte elements
    - hash(bytes) -- 32-bit FNV-1a hash
    - bytes in bytes -- substring test
    - int in bytes -- element membership test
    
Updates https://github.com/bazelbuild/starlark/issues/112

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -103,6 +103,7 @@ interact with the environment.
     * [any](#any)
     * [all](#all)
     * [bool](#bool)
+    * [bytes](#bytes)
 <!--    * [chr](#chr) -->
     * [dict](#dict)
     * [dir](#dir)
@@ -129,6 +130,7 @@ interact with the environment.
     * [type](#type)
     * [zip](#zip)
   * [Built-in methods](#built-in-methods)
+    * [bytes·elems](#bytes·elems)
     * [dict·clear](#dict·clear)
     * [dict·get](#dict·get)
     * [dict·items](#dict·items)
@@ -147,10 +149,10 @@ interact with the environment.
     * [list·remove](#list·remove)
     * [set·union](#set·union)
     * [string·capitalize](#string·capitalize)
-    * [string·codepoint_ords](#string·codepoint_ords)
-    * [string·codepoints](#string·codepoints)
+<!-- * [string·codepoint_ords](#string·codepoint_ords) -->
+<!-- * [string·codepoints](#string·codepoints) -->
     * [string·count](#string·count)
-    * [string·elem_ords](#string·elem_ords)
+<!-- * [string·elem_ords](#string·elem_ords) -->
     * [string·elems](#string·elems)
     * [string·endswith](#string·endswith)
     * [string·find](#string·find)
@@ -810,23 +812,24 @@ The slice expression `b[i:j]` returns the subsequence of `b`
 from index `i` up to but not including index `j`.
 The index expression `b[i]` returns the int value of the ith element.
 
-Like strings, bytes are hashable, totally ordered, and not iterable,
+The `in` operator may be used to test for the presence of one bytes
+as a subsequence of another, or for the presence of a single `int` byte value.
+
+Like strings, bytes values are hashable, totally ordered, and not iterable,
 and are considered True if they are non-empty. 
 
+A bytes value has these methods:
+
+* [`elems`](#bytes·elems)
 ```
 TODO(https://github.com/bazelbuild/starlark/issues/112)
-- methods. Likely the same as string (minus those concerned with text):
-    elems - iterator over ints
+- more methods: likely the same as string (minus those concerned with text):
     join
     {start,end}with
     {r,}{find,index,partition,split,strip}
     replace
-- specify ord, chr?
-- hash(bytes)
-- support 'bytes in bytes', 'int in bytes'?
-- bytes(...) function
-- encode, decode methods?
-- can we reduce string iterator methods without loss of generality/efficiency?
+TODO: encode, decode methods?
+TODO: ord, chr.
 ```
 
 ### Lists
@@ -1271,10 +1274,10 @@ Its [type](#type) is `"builtin_function_or_method"`.
 
 A built-in function value used in a Boolean context is always considered true.
 
-Many built-in functions are predeclared in the environment
-(see [Name Resolution](#name-resolution)), and are thus available to	(see [Name Resolution](#name-resolution)).
-all Skylark programs.	Some built-in functions such as `len` are _universal_, that is,
-available to all Skylark programs.
+Many built-in functions are predeclared in the environment;
+see [Name Resolution](#name-resolution).
+Some built-in functions such as `len` are _universal_, that is,
+available to all Starlark programs.
 The host application may predeclare additional built-in functions
 in the environment of a specific module.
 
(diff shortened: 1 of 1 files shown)


alandonovan:
Some comments on what is (and what is not) in this PR:

- There is still no spec for `ord` nor `chr` functions. 
 
- `chr` is easy enough to define: it takes an int, and returns a 1-codepoint string. If you want bytes of the result, call `bytes(chr(x))`, for now.

- codepoint_chr is not needed. Use chr.

- Python's `ord` function returns the numeric value of a 1-element string. (Python2 string elements are bytes; Python3 strings are of code points, and Python3 bytes strings of bytes.) Given that Starlark's text string elements are not code points, I think it risks confusion for us to define `ord`.  One can use `bytes.elems()[0]` to obtain the same result, at some constant overhead, and this works on strings longer than 1 as well.  If `string.elem_ords()` existed in the spec (and not just the Go impl), you could use that too. I think we should specify a method like elem_ords for strings, but ideally it would be called string.elems, for symmetry with bytes.elems. Unfortunately, that method already exists and returns 1-element substrings, which is really not useful (and would be an error for non-ASCII in Rust.) I propose we use elem_ords for now, but plan longer term to make elems equivalent to elem_ords.

- codepoint_ord(s) can be achieved using `s.codepoint_ords()[0]`, at least for now. We should specify string.codepoint_ords, which only exists in the Go impl. (Tricky to implement in Java due to Bazel latin1 hack.)

- bytes(str) and str(bytes): this PR defines these conversions, with U+FFFD behavior, not fail-fast. Why do we not want to support these operations, as comments in #112 say?

- If we ever want to support other encodings (and I think that's unlikely), we needn't pollute the str() and bytes() functions with extra parameters. We could introduce explicit encode/decode methods in that case.

spec.md
```diff
@@ -3107,23 +3138,21 @@ provided `default` value instead of failing.
...
 ### hash
 
-`hash(x)` returns an integer hash of a string x
-such that two equal strings have the same hash.
+`hash(x)` returns an integer hash of a string or bytes x
+such that two equal values have the same hash.
 In other words `x == y` implies `hash(x) == hash(y)`.
+Any other type of argument in an error, even if it is suitable as the key of a dict.
+
 In the interests of reproducibility of Starlark program behavior over time and
```
brandjon, line 3146: > In the interests of reproducibility

Which is exactly what Python aims to avoid for the sake of thwarting collision DoS attacks... Can't be helped, I suppose. Do we have the option to extend `hash()` with a seed argument in the future? Or perhaps at that point you'd just run the interpreter in a mode where it's allowed to substitute an arbitrary hash function.

spec.md
```diff
@@ -3107,23 +3138,21 @@ provided `default` value instead of failing.
...
-across implementations, the specific hash function is the same as that implemented by
+across implementations, the specific hash function for bytes is 32-bit FNV-1a,
```
brandjon, line 3147: What's the merit of FNV-1a in particular? The merit of matching Java for strings is obvious.

spec.md
```diff
@@ -3326,9 +3355,13 @@ str(1)                          # '1'
 str("x")                        # 'x'
 str([1, "x"])                   # '[1, "x"]'
 str(0.0)                        # '0.0'        (formatted as if by "%g")
-str(b"abc")                     # 'b"abc"'
+str(b"abc")                     # 'abc'
 ```
 
+The string form of a bytes value is the UTF-K decoding of the bytes.
+Each byte that is not part of a valid encoding is replaced by the
```
brandjon, line 3362: Same issue as above regarding the replacement algorithm.

brandjon (approved):

spec.md
```diff
@@ -1272,8 +1275,9 @@ Its [type](#type) is `"builtin_function_or_method"`.
 A built-in function value used in a Boolean context is always considered true.
 
 Many built-in functions are predeclared in the environment
-(see [Name Resolution](#name-resolution)), and are thus available to	(see [Name Resolution](#name-resolution)).
-all Skylark programs.	Some built-in functions such as `len` are _universal_, that is,
+(see [Name Resolution](#name-resolution)),
+and are thus available to all Skylark programs.
```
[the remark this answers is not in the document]
alandonovan: Done.

spec.md
```diff
@@ -1272,8 +1275,9 @@ Its [type](#type) is `"builtin_function_or_method"`.
 A built-in function value used in a Boolean context is always considered true.
 
 Many built-in functions are predeclared in the environment
```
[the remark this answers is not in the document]
alandonovan: There was a preexisting error in the text here. Fixed to:
```
Many built-in functions are predeclared in the environment; see [Name Resolution](#name-resolution).                                                                                      
Some built-in functions such as `len` are _universal_, that is, available to all Starlark programs.                                                                                           
The host application may predeclare additional built-in functions                                                             
in the environment of a specific module.         
```

spec.md
```diff
@@ -2110,19 +2114,20 @@ these operators.
 #### Membership tests
 
 ```text
-      any in     sequence		(list, tuple, dict, string)
+      any in     sequence		(list, tuple, dict, string, bytes)
```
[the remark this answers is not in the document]
alandonovan: Ideally the spec would get into the business of exhaustively listing all possible types, instead referring to abstractions like "indexable sequence" and given examples of one or to instances.

I've added range here and in a couple places, but it's a slippery slope.

spec.md
```diff
@@ -2984,6 +2992,29 @@ If the iterable is empty, it returns `True`.
...
 With no argument, `bool()` returns `False`.
 
+### bytes
+
+`bytes(x)` converts its argument to a `bytes`.
+
+If x is a `bytes`, the result is x.
+
+If x is a string, the result is a `bytes` whose elements are
+the UTF-8 encoding of the string. Each element of the string that is
```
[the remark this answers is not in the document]
alandonovan: Yes, the Unicode spec doesn't prescribe it exactly, and this question has a history; see https://hsivonen.fi/broken-utf-8/Unicode. But it's safer for us to prescribe the behavior here, as that very document points out:

> Web standards these days tend to avoid implementation-defined behavior due to the painful experience of Web sites developing dependencies on the quirks of particular browsers in areas that old specs considered error situations not worth spelling out precise processing requirements for. Therefore, there has been a long push towards well-defined behavior even in error situations on the Web without debating each particular error case...

I chose this behavior because it's easy to implement, and it's the one required by the Go spec, and if I recall correctly, Java's String.getBytes(utf8) uses '?' as its replacement (!!), and thus isn't compatible with any reading of Unicode, so we'll have to handle the conversion ourselves anyway.

spec.md
alandonovan: Yes, but Starlark prizes determinism higher than DoS protection. This was a material problem when I implemented Herb: execution of BUILD files would vary based on the hash function, causing (for example) different elements to be chosen from a set, causing different results, and in some cases, errors only in one implementation.

> Do we have the option to extend hash() with a seed argument in the future? Or perhaps at that point you'd just run the interpreter in a mode where it's allowed to substitute an arbitrary hash function.

No, we don't have an option.

spec.md
alandonovan: It's extremely simple and good enough.

spec.md
alandonovan: Yep, same answer.

spec.md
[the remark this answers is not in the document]
brandjon: If this is consistent with practices in Go, it's good enough for me.

spec.md
brandjon: Well, put it this way: If Starlark is ever used by an application that wants DoS protection and values it over determinism, they'll do what they want in the implementation (to the extent that they control their interpreter) and ignore the spec. So it's just a choice of whether we bless it or not, and I don't feel strongly either way.

spec.md
brandjon: Sure, was just wondering if it was used in Go, or why -1a vs -1 . I'll trust your pick though.

alandonovan: It's what I had already implemented. According to wikipedia, doing the XOR before the MUL leads to "slightly better avalanche characteristics".

PR #167, opened 2021-02-14
Fix formatting of table of contents

brandjon:
HTML comments apparently don't work within a list. Replaced them with empty links.


laurentlb:
These links are not valid though.
As they break the markdown, we could remove the lines?

brandjon:
Done, pushed them into TODO comments instead.

laurentlb (approved):
lgtm, thanks

PR #207, opened 2021-10-06
Use main branch for starlark-rust repo

styurin:
https://github.com/facebookexperimental/starlark-rust/commits/main

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -90,7 +90,7 @@ go_repository(
 new_git_repository(
   name = "starlark-rust",
   remote = "https://github.com/facebookexperimental/starlark-rust.git",
-  branch = "master",
+  branch = "main",
   build_file_content = """
 genrule(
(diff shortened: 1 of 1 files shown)


laurentlb:
ping @brandjon

laurentlb:
I should be able to merge when the tests pass. We could comment out the tests that don't pass if needed.

fmeum:
The remaining CI failures are all due to Rust Starlark and seem to be about error message patterns. However, `bazel test //...` passes on my local machine, including the tests that fail in the CI, which is very weird.

fmeum:
@ndmitchell Do you happen to have an idea what's wrong with the tests? They pass locally (at least on my machine).

laurentlb:
It's probably an old error - I've merged https://github.com/bazelbuild/starlark/pull/213 instead, as the tests were passing.

PR #212, opened 2022-02-15
Add removeprefix/removesuffix to string

### LINKED ISSUE #185: Add removeprefix/removesuffix to string

fmeum:
The names and behavior of the new functions `removeprefix` and `removesuffix` are consistent with [PEP 616](https://www.python.org/dev/peps/pep-0616/).

Fixes #185.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -164,6 +164,8 @@ interact with the environment.
     * [string·lower](#string·lower)
     * [string·lstrip](#string·lstrip)
     * [string·partition](#string·partition)
+    * [string·removeprefix](#string·removeprefix)
+    * [string·removesuffix](#string·removesuffix)
     * [string·replace](#string·replace)
     * [string·rfind](#string·rfind)
     * [string·rindex](#string·rindex)
@@ -3924,6 +3926,30 @@ If S does not contain `x`, `partition` returns `(S, "", "")`.
 
 ```python
 "one/two/three".partition("/")		# ("one", "/", "two/three")
+```
+
+<a id='string·removeprefix'></a>
+### string·removeprefix
+
+`S.removeprefix(prefix)` returns a copy of string S with the prefix `prefix`
+removed if S starts with `prefix`, otherwise it returns S.
+
+```python
+"banana".removeprefix("ban")          # "ana"
+"banana".removeprefix("foo")          # "banana"
+"foofoobar".removeprefix("foo")       # "foobar"
+```
+
+<a id='string·removesuffix'></a>
+### string·removesuffix
+
+`S.removesuffix(suffix)` returns a copy of string S with the suffix `suffix`
+removed if S ends with `suffix`, otherwise it returns S.
+
+```python
+"banana".removesuffix("nana")         # "ba"
+"banana".removesuffix("foo")          # "banana"
+"banana".removesuffix("na")           # "bana"
 ```
 
 <a id='string·replace'></a>


fmeum:
The CI fails with with an errors that seems unrelated:

```
Error in fail: error running 'git fetch origin master:origin/master' while working with @starlark-rust:
fatal: Couldn't find remote ref master
```

spec.md
```diff
@@ -3926,6 +3928,30 @@ If S does not contain `x`, `partition` returns `(S, "", "")`.
...
+<a id='string·removeprefix'></a>
+### string·removeprefix
+
+`S.removeprefix(prefix)` returns a copy of string S with the prefix `prefix`
+removed if S starts with `prefix`, otherwise it returns S.
+
+```python
+"banana".removeprefix("ban")          # "ana"
+"banana".removeprefix("foo")          # "banana"
+"foofoobar".removeprefix("foobar")    # "foo"
```
laurentlb, line 3940: This one looks wrong

laurentlb (approved):
The change looks good to me. Is it already implemented in a Starlark interpreter?

spec.md
fmeum: Sorry, got argument and return value mixed up. Should be fixed now.

fmeum:
> The change looks good to me.
> 
> Edit: Implementation in Java: [bazelbuild/bazel#14824](https://github.com/bazelbuild/bazel/pull/14824)

According to https://github.com/bazelbuild/starlark/issues/185#issue-832557499, they are also available in Rust Starlark.

laurentlb:
I think we need to fix the tests (in a separate PR) before we can merge this.

(or we'll need an admin)

adonovan (approved):
LGTM

fmeum:
@adonovan @laurentlb Only just noticed that there is #193, which has more details and added the methods to one more ToC. Should that one be merged instead?

laurentlb:
Done.

Thanks a lot for your help!

tetromino:
The spec change contradicts PEP 616: if the string does not have a prefix/suffix, we want to (as in PEP 616) return a *copy* of the string, not the original string.

Do we really want `foo.removeprefix(".so") += ".dll"` to sometimes mutate `foo` and sometimes not?

adonovan:
> The spec change contradicts PEP 616: if the string does not have a prefix/suffix, we want to (as in PEP 616) return a _copy_ of the string, not the original string.
> 
> Do we really want `foo.removesuffix(".so") += ".dll"` to sometimes mutate `foo` and sometimes not?

Starlark doesn't have a notion of string identity, so you can't distinguish the original from a copy. That said, it would be more precise to express it as: S.removeprefix(prefix) returns the portion of S after the prefix `prefix`, if S starts with that prefix, otherwise it returns S.

laurentlb:
> to sometimes mutate `foo` and sometimes not?

Strings are immutable in Starlark.

PR #216, opened 2022-02-17
Document key parameter of min

laurentlb:
Similar to the description in `max`


tetromino (approved):
Seems like an obvious commission. Sadly neither `max` nor `min` in Bazel support the parameter - we ought to fix that...

tetromino:
Made an issue to implement this in Bazel: https://github.com/bazelbuild/bazel/issues/15022

PR #229, opened 2022-08-25
added missing 'x' in hexadecimal escape example

aeldidi:
The spec says that `\0A` as an escape sequence should evaluate to the newline character (`0x0A` in hex),
when it is actually octal 0 followed by the character 'A'. I believe the spec meant to say `\x0A`, which does
what is described.


adonovan (approved):
Thanks!

laurentlb (approved):

PR #234, opened 2022-11-14
Add abs function

ndmitchell:
The `abs` function is simple, and available in Python. Having it available in Starlark would be useful - it is already available in the starlark-rust library.


laurentlb (approved):
Looks good to me

PR #240, opened 2023-01-11
Point to newer git commit of Java starlark interpreter; add and update deps for java and go

### LINKED ISSUE #238: Tests are failing
e.g. https://buildkite.com/bazel/starlark/builds/1560#01856489-126f-4ce9-a4df-e6c800c59a26

> (19:37:06) ERROR: /workdir/test_suite/BUILD:5:16: in py_test rule //test_suite:starlark_java_test_testdata/java/int_star: target '@io_bazel//src/main/java/net/starlark/java/cmd:starlark' is not visible from target '//test_suite:starlark_java_test_testdata/java/int_star'. Check the visibility declaration of the former target if you think the dependency is legitimate

@tetromino @brandjon, could you take a look?
Tests started to fail between March and October (https://buildkite.com/bazel/starlark), I don't know if anything changed on the Bazel side. This prevents us from merging changes in the repository.

tetromino:
Needed to unbreak tests with Bazel 6.0

Fixes #238

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -45,8 +45,40 @@ rules_pkg_dependencies()
 
 git_repository(
     name = "io_bazel",
+    commit = "4853dfd02ac7440a04caada830b7b61b6081bdfe", # 2023-01-11
     remote = "https://github.com/bazelbuild/bazel.git",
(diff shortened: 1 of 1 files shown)


brandjon (approved):
<Rubber stamp>

PR #248, opened 2023-03-16
Add keyword-only arguments to the spec

### LINKED ISSUE #23: Keyword-only arguments
Keyword-only arguments are described in a Python PEP: https://www.python.org/dev/peps/pep-3102/

Examples:

```python
def compare(a, b, *, key=None): ...

def sortwords(*wordlist, case_sensitive=False): ...
```

It is supported in the Java implementation. There was a desire to remove this feature, to keep Starlark a subset of Python 2 and Python 3 syntax. However, this feature is useful and often requested:
 * Many builtins use it
 * Documentation in Bazel uses that syntax
 * Macros in Bazel shouldn't allow positional arguments, for maintenance reasons

This doesn't conflict with https://github.com/bazelbuild/starlark/issues/13#issuecomment-433995990, because #13 concerns only the call sites. The current issue concerns only the definition sites.

laurentlb:
The wording is copied from https://github.com/google/starlark-go/pull/143 and https://github.com/google/starlark-go/pull/151

Fixes #23

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2721,7 +2721,7 @@ def g(a, *args, b=2, c):
 g(1, 3)                 # error: function g missing 1 argument (c)
 g(1, *[4, 5], c=3)      # error: keyword argument c may not follow *args
 g(1, 4, c=3)            # "1 2 3 (4,)"
-g(1, c=3, *[4, 5])      # "1 2 3 (4,5)"
+g(1, c=3, *[4, 5])      # "1 2 3 (4, 5)"
 ```
 
 A non-variadic function may also declare keyword-only parameters,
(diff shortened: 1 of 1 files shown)


tetromino:
> This line is already in the spec.
> 
> https://github.com/bazelbuild/starlark/blob/e2175b4ec1312c196347cef439b36980b06b195f/spec.md?plain=1#L1233

Sorry for being unclear: I meant that the line is now likely to confuse the reader. I pushed a commit to explicitly point out the difference between the use of keyword-only arguments in a function call vs. function declaration.

tetromino (approved):

PR #263, opened 2023-08-18
add `envd` to the users

kemingy:


brandjon:
Sorry for the delay on this. We're tracking users on Laurent's awesome-starlark page, which you're already merged into, so I'll close this now.

PR #268, opened 2023-10-20
add Kurtosis as a project that uses Starlark

leeederek:
Kurtosis is developer tool for Platform Engineers and DevOps teams to define and spin up complex distributed systems for dev, test, and prod. Kurtosis uses Starlark as the DSL for describing how the environment should be stitched together.

PR #269, opened 2023-10-31
add Kurtosis as a project that uses Starlark

galenmarchetti:
This PR should take precedence over #268 , which will be closed shortly. Submitting it via my github username so that the CLA checks pass + I can debug them!

We at Kurtosis use starlark heavily as a language for defining the set up logic of multi-container environments for backend engineers, across dev, test, and prod. Our end users use our DSL, exposed in starlark, to define their environments!


laurentlb:
It looks like I have no longer write access here, so I'll let @brandjon and @tetromino merge the PR.

PR #283, opened 2024-07-31
Clarify that predeclared names can be redefined in the code.

### LINKED ISSUE #282: Spec incorrectly claims that predeclared bindings cannot be reassigned
In the [Name binding and variables](https://github.com/bazelbuild/starlark/blob/master/spec.md#name-binding-and-variables) section, it says:

> The set of predeclared names includes the universal constant values `None`, `True`, and `False`, and various built-in functions such as len and list; these functions are immutable and stateless. [...] Starlark programs cannot change the set of predeclared bindings or assign new values to them.

However both Bazel and https://github.com/google/starlark-go allow reassignment of `True`:

```python
True = 3
print(True)
```

I initially considered this to be a bug in the implementations rather than the spec, since it seems like a good idea to prevent rebinding of the predeclared names. But then I realized that this would make it impossible to add new predeclared names in a backwards-compatible way. So perhaps this is a necessary evil.

laurentlb:
Fix #282


brandjon:
Thanks for the clarification. FYI I opened #312 to restrict the ability to bind to the names `None`, `False`, and `True` specifically.

brandjon (approved):

PR #284, opened 2024-07-31
Fix Go compilation

laurentlb:
Before this commit, the compilation fails with:

```
(18:39:29) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/io_bazel_rules_go/BUILD.bazel:71:16: in (an implicit dependency) attribute of go_context_data rule @@io_bazel_rules_go//:go_context_data: in $whitelist_function_transition attribute of go_context_data rule @@io_bazel_rules_go//:go_context_data: package group '@@bazel_tools//tools/whitelists/function_transition_whitelist:function_transition_whitelist' is misplaced here (they are only allowed in the visibility attribute)
```

(https://buildkite.com/bazel/starlark/builds/2181#01910a16-5592-4949-8b24-6f6dbf212e6f)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -86,10 +86,10 @@ maven_install(
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "56d8c5a5c91e1af73eca71a6fab2ced959b67c86d12ba37feedb0a2dfea441a6",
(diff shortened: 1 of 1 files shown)


laurentlb:
@tetromino

PR #287, opened 2024-10-29
Add recur in `users.md`

dbohdan:
This PR adds my project [recur](https://github.com/dbohdan/recur) to the list of users. recur is a tool for retrying commands that uses Starlark in a small way: to let the user write a [success condition](https://github.com/dbohdan/recur?tab=readme-ov-file#conditions) for the command. No worries if this isn't enough to include it.

Some background: I have migrated recur from Python and [simpleeval](https://github.com/danthedeckie/simpleeval) to Go and starlark-go. I considered the Common Expression Language but didn't want to change the syntax. I am happy with my choice so far. The user-facing changes with starlark-go are relatively minor. The most noticeable are the lack `is None` and having to replace `code in {1, 3}` with `code in (1, 3)`. Starlark expressions seem a viable alternative to CEL for this kind of task.


dbohdan:
Added in [laurentlb/awesome-starlark](https://github.com/laurentlb/awesome-starlark). Closing.

Issue #288, opened 2024-10-29
"Awesome Starlark"


dbohdan:
It could help promote the language and projects that implement, extend, and use it if `users.md` were transformed into an official ["awesome list"](https://github.com/sindresorhus/awesome) repository. A link in sindresorhus/awesome would introduce users from outside the Bazel community to Starlark.

laurentlb:
Interesting idea. users.md doesn't fit very well with the rest of the repository and it can be hard to get maintainers' attention for the small PRs.

I have created https://github.com/laurentlb/awesome-starlark and we can iterate for there.

dbohdan:
Excellent!

dbohdan:
I think it would make sense to replace the contents of `users.md` with a link to Awesome Starlark in order to have a single source of truth.

PR #289, opened 2024-10-29
Remove users.md, it was copied to a separate repository

laurentlb:
Motivation:

  This starlark repository is owned by the BazelBuild team, for which the
  maintenance of this file is not a priority. users.md is not really related
  to Bazel, and pull requests tend to be ignored.

  I believe we can easily add maintainers for this file, as it's not critical.
  By contrast, changes to the spec should continue to be restricted.

  I think this can help create more visibility to Starlark users, and it will
  be easier to evolve and expand the content of the document.

https://github.com/bazelbuild/starlark/issues/288


laurentlb:
@tetromino

brandjon:
Sounds reasonable to me that this repository not be in the business of directly tracking interesting use cases of Starlark. Thoughts Sasha?

tetromino:
I would prefer to at least keep the section about the 3 full language implementations here. (The rest can move.)

tetromino (approved):

PR #297, opened 2025-01-16
Fix typo

lgalfaso:
_function_ blocks and _comprehension_ blocks for top level functions and comprehensions are contained only within the _file_ block, and not directly to the _module_ block.


brandjon:
Thanks!

brandjon (approved):

brandjon:
Looking at the blame, this was probably a merge error from the starlark-go version of the spec.

Echalon0ne (approved):

PR #299, opened 2025-03-03
Remove old out-of-scope proposals

comius:
The proposals are quite old and according to [process.md](https://github.com/bazelbuild/starlark/blob/master/process.md) also out of scope. ToolchainInfo and genrules were never part of Starlark and the proposals should have been published github.com/bazelbuild/proposals.


brandjon:
> according to [process.md](https://github.com/bazelbuild/starlark/blob/master/process.md) also out of scope

Ug, process.md itself is quite out of date. It's describing non-rule-specific .bzl logic as in-scope for this repo, which is a historic artifact. (No action item for this PR though.)

brandjon (approved):

PR #302, opened 2025-03-04
Order fail and float in TOC

comius:


brandjon:
FYI we have a Python script (`spec_md_gen.py`) that updates the table of contents. We haven't been using it for quite a while, but when we fix the CI it'll probably have a backlog of suggestions to make.

brandjon (approved):

PR #303, opened 2025-03-05
Add productions for type annotations.

comius:
Going with the stricter version of productions in comparison with Python.

Pros:
- makes evaluation simpler, there's a lot less to validate
- removes ambiguity between `Dict[(str,str)]==Dict[str,str]` and `List[(int,)] != List[int]`
- stricter syntax can be relaxed later

Cons:
- extends the Parser for additional 100 lines of code 

Issue: https://github.com/bazelbuild/bazel/issues/24407

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -4722,7 +4722,7 @@ Type = identifier [TypeArguments].
 
 TypeArguments = '[' TypeArgument {',' TypeArgument} ']'.
 
-TypeArgument = Type
+TypeArgument = TypeExpr
              | ListOfTypes
              | DictOfTypes
              | string
--- a/spec.md
+++ b/spec.md
@@ -2852,9 +2852,8 @@ type, a colon, and then an indented block of statements which form the body of
 the function.
 
 The parameter list is a comma-separated list whose elements are of
-several kinds. Any parameter name may be associated with a type expression.
-First come zero or more required parameters, which are simple identifiers;
-all calls must provide an argument value for these parameters.
+several kinds.  First come zero or more required parameters, which are
+simple identifiers; all calls must provide an argument value for these parameters.
 
 The required parameters are followed by zero or more optional
 parameters, of the form `name=expression`. The expression specifies
@@ -2871,6 +2870,8 @@ parameters, again of the forms `name` or `name=expression`,
 but these parameters differ from earlier ones in that they are
 _keyword-only_: if a call provides their values, it must do so as
 keyword arguments, not positional ones.
+
+Any parameter name may be associated with a type expression.
 
 Note that even though keyword-only arguments are declared after `*args` in a
 function's definition, they nevertheless must appear before `*args` in a call
--- a/spec.md
+++ b/spec.md
@@ -4725,6 +4725,7 @@ TypeArguments = '[' TypeArgument {',' TypeArgument} ']'.
 TypeArgument = Type
              | ListOfTypes
              | DictOfTypes
+             | string
              .
 
 ListOfTypes = '[' [TypeArgument {',' TypeArgument} [',']] ']'.
--- a/spec.md
+++ b/spec.md
@@ -2852,14 +2852,14 @@ type, a colon, and then an indented block of statements which form the body of
 the function.
 
 The parameter list is a comma-separated list whose elements are of
-several kinds.  First come zero or more required parameters, which are
-simple identifiers with optionally a type annotation; 
+several kinds. Any parameter name may be associated with a type expression.
+First come zero or more required parameters, which are simple identifiers;
 all calls must provide an argument value for these parameters.
 
 The required parameters are followed by zero or more optional
-parameters, of the form `name = expression` or `name: type = expression`.
-The expression specifies the default value for the parameter for use in calls 
-that do not provide an argument value for it.
+parameters, of the form `name=expression`. The expression specifies
+the default value for the parameter for use in calls that do not
+provide an argument value for it.
 
 The required parameters are optionally followed by a single parameter
 name preceded by a `*`.  This is the called the _varargs_ parameter,
@@ -4716,9 +4716,9 @@ LambdaParameter  = identifier ['=' Test]
                  | '**' identifier
                  .
            
-TypeExpr = Type {'|' Type}.
+TypeExpr = TypeAtom {'|' TypeAtom}.
 
-Type = identifier [TypeArguments].
+TypeAtom = identifier [TypeArguments].
 
 TypeArguments = '[' TypeArgument {',' TypeArgument} ']'.
 
--- a/spec.md
+++ b/spec.md
@@ -23,7 +23,7 @@ now maintained by the Bazel team.
 
 ## Overview
 
-Starlark is an untyped dynamic language with high-level data types,
+Starlark is a typed dynamic language with high-level data types,
 first-class functions with lexical scope, and automatic memory
 management or _garbage collection_.
 
@@ -254,7 +254,7 @@ characters are tokens:
 ```text
 +    -    *    /    //   %    **
 ~    &    |    ^    <<   >>
-.    ,    =    ;    :
+.    ,    =    ;    :    ->
 (    )    [    ]    {    }
 <    >    >=   <=   ==   !=
 +=   -=   *=   /=   //=  %=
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -4786,7 +4812,7 @@ Binop = 'or'
       | '*' | '%' | '/' | '//'
       .
 
-LambdaExpr = 'lambda' [Parameters] ':' Test .
+LambdaExpr = 'lambda' [LambdaParameters] ':' Test .
```
brandjon, line 4815: Support an optional trailing comma, like `def` does.

spec.md
```diff
@@ -4695,17 +4698,40 @@ File = {Statement | newline} eof .
...
+
+LambdaParameter  = identifier ['=' Test]
+                 | '*'
+                 | '*' identifier
+                 | '**' identifier
+                 .
+           
+TypeExpr = Type {'|' Type}.
+
+Type = identifier [TypeArguments].
```
brandjon, line 4721: Nit: Use a more specific name so as not to confuse with the toplevel production (`TypeExpr`). Maybe `TypeAtom`?

spec.md
```diff
@@ -4695,17 +4698,40 @@ File = {Statement | newline} eof .
...
+
+TypeArgument = TypeExpr
+             | ListOfTypes
+             | DictOfTypes
+             | string
+             .
+
+ListOfTypes = '[' [TypeArgument {',' TypeArgument} [',']] ']'.
+DictOfTypes = '{' [TypeEntry {',' TypeEntry} [',']] '}' .
+TypeEntry    = string ':' TypeArgument .
```
brandjon, line 4733: Is the key a string or identifier? Is this intended to match a struct field name / record key / dict key?

spec.md
```diff
@@ -4695,17 +4698,40 @@ File = {Statement | newline} eof .
...
+
+TypeArguments = '[' TypeArgument {',' TypeArgument} ']'.
+
+TypeArgument = TypeExpr
+             | ListOfTypes
+             | DictOfTypes
+             | string
+             .
+
+ListOfTypes = '[' [TypeArgument {',' TypeArgument} [',']] ']'.
```
brandjon, line 4731: I don't have a good sense of what `ListOfTypes` and `DictOfTypes` are for.

spec.md
brandjon, line 4721: Is there support for a syntactic sugaring of List[Int] or Tuple[String, Int] as [Int] and (String, Int) respectively?

brandjon (approved):

brandjon:
> Dict[(str,str)]==Dict[str,str] and List[(int,)] != List[int]

Good catch. I agree we shouldn't distinguish between these semantically, as in the case of `List[(int,)]`. Furthermore, since no user code is introspecting these, the fact that `A[B]` parses differently between type expressions and value expressions isn't directly observable to the user. So it doesn't hurt to be conservative/opinionated about syntax.

spec.md
brandjon: Ah, it's clearer reading the prototype implementation CL. `ListOfTypes` is useful for param lists in `Callable`, and `DictOfTypes` is useful for `TypedDict`, the type of a heterogeneous dictionary. Are other uses imagined?

spec.md
```diff
@@ -4695,17 +4698,40 @@ File = {Statement | newline} eof .
...
 
+LambdaParameters = LambdaParameter {',' LambdaParameter}.
+
+LambdaParameter  = identifier ['=' Test]
+                 | '*'
+                 | '*' identifier
+                 | '**' identifier
+                 .
+           
+TypeExpr = Type {'|' Type}.
```
brandjon, line 4719: I take it generics of unions, like `List[Noneable[String]]`, will come later? Or like we have in rules' declared providers param, `List[Provider | List[Provider]]`.

For `Noneable`, you could imagine special casing it in the type system if you didn't want union inside generics.

spec.md
brandjon: Ack. Filed #304.

spec.md
```diff
@@ -4695,17 +4698,40 @@ File = {Statement | newline} eof .
...
 
 Parameters = Parameter {',' Parameter}.
 
-Parameter  = identifier
-           | identifier '=' Test
+Parameter  = identifier [':' TypeExpr] ['=' Test]
            | '*'
-           | '*' identifier
-           | '**' identifier
+           | '*' identifier [':' TypeExpr]
```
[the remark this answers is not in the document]
brandjon: FYI some of the examples I've seen in Python suggest that I had it wrong, and `*args`  would be annotated with the element type, e.g. `Int` rather than `Tuple[Int]`.

spec.md
brandjon: This isn't to say don't prototype it, but...

I don't think users would like to read/write callable types as in your example syntax `fun[...]`, with separate sections delineating which params can be passed in which way. We might have to use a more verbose notation (with multiple lines) for declaring such a type and binding it to a type alias for convenient reference in signatures.

Looking at your example, there's nothing in the syntax that would clue me into the fact that the different generic positions are for different types of argument passing. Maybe an even more wordy form like `fun[positional_only=[int], positional_or_keyword=...]` would do it, but you can see how terrible that would be to read and write.

I'd much prefer your more concise syntax that isn't supported in Python syntax. Though honestly, even that may still be a terror to read in practice for any non-trivial function type.

spec.md
brandjon: I thought Python had this but I can't find it now so probably not.

I'm tempted to say we should add it, but I fear it'd be useless if we (or users?) subtype these builtin types with ones that have more invariants, e.g. FrozenDict or whatnot.

spec.md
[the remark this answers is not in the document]
comius: I know about that. I think Python had it wrong, not you :)

spec.md
comius: Opt1: I think the most common case that users need are keyword arguments. For improved readability, we could just support that variant, i.e. `fun[{"c": float, "d": [List[int]]}, bool]`.

Opt2: Maybe a tad better general option: `fun[[int], {"a": int, "b": [str], "*": None, "c": float, "d": [List[int]]}, bool]`. This way the `list` and the `dict` signify positional and keyword args and `*` in the middle is self-explanatory (I wish - even for regular functions people are asking why is the star there).

comius: Nah, I take back Opt2, because of `fun[{"*": None, "c": float, "d": [List[int]]}, bool]` needed for kw only.

spec.md
comius: Yes that. Maybe you remember Buck having syntax sugar. They removed it. I think, because it was too confusing.

spec.md
comius: I don't follow the comment. With this syntax, `|` is for unions. `List[None | String]` for example. I'm slightly leaning towards this notation (beauty is in the eyes of the beholder, I know).

It would simplify the grammar a lot if I just dropped `|` and go with `List[Noneable[String]]` or `List[Optional[String]]` and `Union[A,B]`.

spec.md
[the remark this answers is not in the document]
brandjon: Counterargument: For `*args`, the only useful subtype of `List[Any]` is `List[<Texpr>]`, not something like `FrozenList[...]`. Fighting for fewer chars to type isn't so bad.

Counter-counterargument: The magic syntax doesn't generalize well to `**kwargs` (which is more common, no?), even for homogeneously typed dicts. Maybe you can do it with list-type-expressions without a preceding type name (i.e. `[K, V]`). And even if it did, magic is still magic.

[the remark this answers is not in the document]
brandjon: (Non-blocking, feel free to resolve)

spec.md
brandjon: It's a little confusing to have order be significant for something in `{}`, even though it's a type expression and not a normal dictionary. I do like matching normal declaration syntax when we can though, rather than overly schemafying what people have to write in a type expression.

spec.md
brandjon: For some reason I had read that grammar as saying that `|` was only permitted at the top level and not in generics. Nevermind.

brandjon (approved):

brandjon:
I attempted to merge the CI-fixing PRs before this one on principle, but was defeated by GH running the check against your branch rather than the merged result. Force-merging for hopefully the last time...

spec.md
brandjon: > * Fix TypeArgument to evolve to TypeExpr instead of Type.

Oh, that's probably why I thought that. ;)

PR #309, opened 2025-03-13
Update grammar to match terminology

comius:
There's no TypeAtom, just Type. First production of Type is a simple type, second production is a constructor applied to other types, resulting in a type.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2847,8 +2847,8 @@ twice("two")            # "twotwo"
 ```
 
 The function's name is preceded by the `def` keyword and followed by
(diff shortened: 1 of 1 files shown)


brandjon (approved):
Thanks for improvement.

I thought you had also added `{}` and `|` to type arguments but maybe that was a different version, or else just in the prototype implementation.

brandjon:
Just realized I didn't merge this although I approved it. I played around with the branch protection settings and created a rule to treat `starlark-with-types` similarly to `master`, but added you with write access to this branch. Let's test it out by having you hit squash-and-merge; let me know if it didn't work.

PR #313, opened 2025-03-28
Commit .bazelversion and MODULE.bazel.lock files

brandjon:
@Wyverald informs me it's best practice to include these files in the repo for reproducibility. Also updated the checked-in.gitignore.


.bazelversion
```diff
@@ -0,0 +1 @@
+8.1.1
```
albertocavalcante, line 1: maybe now also set to 8.4.0 (if works)?

tetromino: Agree - 8.1.1 is pretty old.

tetromino (changes requested):

PR #315, opened 2025-05-16
Union types

comius:


spec.md
```diff
@@ -871,6 +878,14 @@ TODO: ord, chr.
 TODO: string.elems(), string.elem_ords(), string.codepoint_ords()
 ```
 
+### Unions
+
+Some functions may accept parameters of multiple types. In this
+case parameter is annotated with a union type `A|B`.
+
+User functions are limited to optional types, that is `None|A`,
```
brandjon, line 886: Nit: "user functions" might be ambiguous between application-defined vs starlark-programmer-defined. Maybe "Starlark-defined functions"? Not sure whether there's precedent terminology elsewhere in this spec.

spec.md
```diff
@@ -3333,6 +3350,8 @@ See https://github.com/bazelbuild/starlark/issues/47.
 
 ### float
 
+`float(x: str|bool|int|float, /) -> float`
```
brandjon, line 3353: (No action needed for this PR) There may be a bunch of other args to other builtins that are Noneable, in the sense that we take `None` as an alternative to omitting the parameter entirely.

comius: There are bunch of None able args. Since we only have primitive types and unions at this point, I wasn't able to type those. But certainly more is coming.

Issue #206, opened 2021-08-31
Clarify Sequence Type Relationships


mahmoudimus:
In the spec, under [`Sequence types`](https://github.com/bazelbuild/starlark/blob/master/spec.md#sequence-types), it states:

> Sequence: a sequence of known length lets us know how many elements it contains without processing them. Examples: dict, list, tuple, but not string or bytes.

However, in the Java implementation, `Dict` does not implement the `Sequence` interface. I think this is a mistake in the spec and should be removed:

```diff
-Sequence: a sequence of known length lets us know how many elements it contains without processing them. Examples: dict, list, tuple, but not string or bytes.
+Sequence: a sequence of known length lets us know how many elements it contains without processing them. Examples: list, tuple, but not string or bytes.
```

This exposes a few other questions that the spec mentions:

- `SetIndexable` currently does not exist. The Java implementation relies on [`EvalUtils#setIndex`](https://github.com/bazelbuild/bazel/blob/958316f7e0c74bc2f54df9ffc02dcb39ff6d8494/src/main/java/net/starlark/java/eval/EvalUtils.java#L473-L491) but is hardcoded to only `Dict` and `List`. Should there be a `StarlarkSetIndexable` interface or should there be a patch to the existing `StarlarkIndexable`, similar to [`Evalutils#setField`](https://github.com/bazelbuild/bazel/blob/958316f7e0c74bc2f54df9ffc02dcb39ff6d8494/src/main/java/net/starlark/java/eval/EvalUtils.java#L494-L500)? For example:

```diff
+  /**
+   * Updates an object as if by the Starlark statement {@code object[key] = value}.
+   *
+   * @throws EvalException if underlying object is immutable.
+   */
+  void setIndex(StarlarkSemantics semantics, Object key, Object value) throws EvalException;
```

- The concept of a `Mapping` is in the spec (as per #198), but does not exist. Should that be introduced alongside with the modification for `SetIndexable` above? For example:

```java
package net.starlark.java.eval;

import java.util.Map;

public interface StarlarkMapping<K, V> extends Map<K, V>, StarlarkSetIndexable, StarlarkIterable<K> {

}
```

This would also require a change in [`Starlark#len()`](https://github.com/bazelbuild/bazel/blob/958316f7e0c74bc2f54df9ffc02dcb39ff6d8494/src/main/java/net/starlark/java/eval/Starlark.java#L269-L270).

[...]

PR #318, opened 2025-08-08
Resolve Iterable, Sequence, Indexable differences

comius:
Having 3 different types causes confusion. Simplifying down to only 2 type: Iterables and Sequences, that can be easily discerned.

Previously `dict` and `sets` were `Sequences` but not `Indexable` (because of int indices). Merge together `Sequence` and `Indexable`.

This also brings specs closer to Java implementation.

Addresses: https://github.com/bazelbuild/starlark/issues/206

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -566,7 +566,7 @@ every value has a type string that can be obtained with the expression
 expression `str(x)`, or to a Boolean truth value using the expression
 `bool(x)`.  Other operations apply only to certain types.  For
 example, the indexing operation `a[i]` works only with strings, bytes values, lists,
-and tuples, and any application-defined types that are _indexable_.
+and tuples, and any application-defined sequences.
 The [_value concepts_](#value-concepts) section explains the groupings of
 types by the operators they support.
 
@@ -875,7 +875,7 @@ TODO: string.elems(), string.elem_ords(), string.codepoint_ords()
 A list is a mutable sequence of values.
 The [type](#type) of a list is `"list"`.
 
-Lists are indexable sequences: the elements of a list may be iterated
+Lists are sequences: the elements of a list may be iterated
 over by `for`-loops, list comprehensions, and various built-in
 functions.
 
@@ -1762,20 +1762,18 @@ contexts dictionaries act like a sequence of their keys.
 We can classify different kinds of sequence types based on the
 operations they support.
 
-* `Iterable`: an _iterable_ value lets us process each of its elements in a fixed order.
+* `Iterable`: an _iterable_ value lets us process itself (its elements or keys) in a fixed order.
   Examples: `dict`, `list`, `tuple`, `set`, but not `string` or `bytes`.
-* `Sequence`: a _sequence of known length_ lets us know how many elements it
-  contains without processing them.
-  Examples: `dict`, `list`, `tuple`, `set`, but not `string` or `bytes`.
-* `Indexable`: an _indexed_ type has a fixed length and provides efficient
+* `Sequence`: a _sequence of fixed length_ lets us know how many elements it
+  contains without processing them and provides efficient
   random access to its elements, which are identified by integer indices.
-  Examples: `string`, `bytes`, `tuple`, and `list`, but not `dict` or `set`.
+  Examples: `list`, `tuple`, `string`, `bytes`, but not `dict` or `set`.
 * `SetIndexable`: a _settable indexed type_ additionally allows us to modify the
   element at a given integer index. Example: `list`.
 * `Mapping`: a mapping is an association of keys to values. Example: `dict`.
 
 Although all of Starlark's core data types for sequences implement at
-least the `Sequence` contract, it's possible for an an application
+least the `Sequence` contract, it's possible for an application
 that embeds the Starlark interpreter to define additional data types
 representing sequences of unknown length that implement only the `Iterable` contract.
 
@@ -2572,7 +2570,7 @@ f("n")                                          # 2
 
 ### Index expressions
 
-An index expression `a[i]` yields the `i`th element of an _indexable_
+An index expression `a[i]` yields the `i`th element of a sequence
 type such as a string, bytes, tuple, list, or range.  The index `i` must be an `int`
 value in the range -`n` ≤ `i` < `n`, where `n` is `len(a)`; any other
 index results in an error.
@@ -2615,7 +2613,7 @@ type, such as a tuple or string, or a frozen value of a mutable type.
 ### Slice expressions
 
 A slice expression `a[start:stop:stride]` yields a new value containing a
-subsequence of `a`, which must be an indexable sequence such as string,
+subsequence of `a`, which must be a sequence such as string,
 bytes, tuple, list, or range.
 
 ```text


comius:
I decided to open this can of worms. Main motivation is to get a nice `Iterable` and `Sequence` type, that I can use for Starlark types.

brandjon:
Yes, let's open and close this can at long last.

From what I can see, the Python docs don't explicitly define the "indexable" concept the way they do "iterable" and "sequence" (e.g. in [`collections`](https://docs.python.org/3/library/collections.abc.html), [`stdtypes`](https://docs.python.org/3/library/stdtypes.html), or the [glossary](https://docs.python.org/3/glossary.html)). They do say that [all sequences are also iterables](https://docs.python.org/3/glossary.html#term-iterable), which doesn't hold in Starlark due to the exclusion for `str`/`bytes` -- or else strings are to be considered something other than sequences.

A small [reddit thread](https://www.reddit.com/r/learnpython/comments/g29jqe/are_there_any_examples_of_python_objects_that_are/) claims that there are no examples of non-iterable indexables. This is borne out by the [spec](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) of the `__getitem__()` method, and I confirmed with a quick check in the REPL. So it sounds like "indexable" is not necessary as a user-level concept except to the extent we want to break out every possible operation as an interface, which I don't think is useful in the language spec.

I do see there's still a `SetIndexable` type after this PR. That looks like an oversight -- remove? If we do need it, let's be consistent with Python's [ontology](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) and call it `MutableSequence`.

What does that leave us with? As of this PR, Starlark has:

* `Iterable`, defined by being able to appear on the RHS of a for loop. Compatible with the Python ABC of the same name.
* `Sequence`, defined as an indexable iterable of determinable length (unless strings are sequences, in which case we drop "iterable").
* `Mapping`, a concept independent of the above two contracts, since we have mappings (like `TransitiveInfoCollection` in Bazel) that have only associative key lookups without the ability to take their size or iterate over their contents without knowing their key ahead of time.

It would be nice if we could preserve compatibility of these concepts with Python, but that's impossible because of `Mapping`, so we may as well not let that constrain us on `Sequence` w.r.t. strings/bytes.

We may also want `Hashable` at some point, but let's ignore that for the purpose of this PR. We probably don't need other concepts listed in Python's ABCs page.

tetromino:
> * `Mapping`, a concept independent of the above two contracts, since we have mappings (like `TransitiveInfoCollection` in Bazel) that have only associative key lookups without the ability to take their size or iterate over their contents without knowing their key ahead of time.

Note that Python's `Mapping` is sized and iterable.

I would preserve `Mapping` for dict-line objects (including sized, iterable immutable views), and use the term `Indexable` for things like `TransitiveInfoCollection` where we may not allow iteration.

In other words - instead of eliminating `Indexable`, change the definition to allow non-integer keys....

comius:
Option 1 by brandjon: Iterable, Sequence, Mapping;  
Option 2 by tetromino: Iterable, Sequence, Mapping, Indexable with non-int keys 

Arguing for Option 1:
We're already saying that Sequence is not necessarily Iterable (because of str).
Why not do similarly and say Mapping is not necessarily Iterable? We're committing the same incompatibility with Python as with Sequence. And we don't need to introduce a separate Indexable type.

brandjon:
After thinking about this a bit more, I have a new proposal that keeps us more consistent with Python while staying minimalistic. The fewer names we have, and the fewer differences from Python, the less pain we all feel down the line. (Imagine writing a type signature of a bunch of unions of mixins, like `def f(x : Iterable[T]|Mapping[T, V]|Sequence[T]|Sized...)`).

One important observation is that Starlark doesn't currently have consumable iterators, nor would a frozen iterable be useful since that would prevent reading its contents. This seems to eliminate much of the need for a separate `Iterable` type.

The other differences between `Iterable` and `Sequence` in Python are that the latter is `Sized` and `Container` and of course indexable. Maybe we can contract some of these futher, e.g. by replacing `Iterable` with `Collection` (= `Iterable` + `Sized` + `Container`)? This would fix a problem in this PR's current wording for `len(x)` that allows `x` to be an iterable or sequence, which doesn't make sense if iterables are not sized -- instead we'd say `x` is a collection or sequence.

Going further, perhaps we don't care to treat strings as sequences at all. Not only are they not iterable, but the membership operation is weird for them (`in` means "is substring of"). That would allow us to say `Sequence` is a subtype of `Collection`, consistent with Python, and we could drop the confusing case of an "iterable sequence" in both prose and union type signatures, at the cost of having to union sequences with strings/bytes in some signatures. Then `len(x)` allows `x` to be either a collection or a string/bytes. Meanwhile, stuff like `any(x)` takes a collection, not an iterable sequence. (BTW the Java implementation, Go implementation, and Python all support `any({True: 1})` while the spec currently does not.) I imagine `enumerate()`, `range()`, and `zip()` would all be specified to return sequences.

This works so long as we don't have (non-destructively consumed) streams of indeterminate length. If an application does define such a type, it wouldn't satisfy the `Collection` contract, though we could still allow the interpreter to consider it iterable via an internal interface -- there's just no characteristic name for that concept in the spec and type system. Users could still write an ad hoc `Union[Sequence[T], WeirdStreamType]`, or just lift the type to `object` or `Any`.

Maybe this is a problem if we want a built-in function like `any()` to take the iterable and still pass a static type check, or if we add Python's `map()` function and want the result to not necessarily be sized (`len(map(...))` is disallowed in Python). But we could also just ban it for simplicity and make users wrap it in a list, or have the application declare compatibility with the type and raise a not-implemented error dynamically.

Finally, since we would now have compatibility with Python on the notions of `Collection` and `Sequence` (modulo strings/bytes), we could define `Mapping` as a subtype of `Collection` (precisely, `Mapping[K, V]` <: `Collection[K]`). This simplifies `dict`'s type inheritance considerably. `TransitiveInfoCollection` would no longer be considered a `Mapping`, but just a funky application type that supports subscripting, analogous to the stream example.

brandjon:
I am wavering a little on whether we should drop `Iterable` or keep it in addition to `Collection`, i.e. have `Sequence` <: `Collection` <: `Iterable` (and likewise subbing `Mapping` for `Sequence`). We could also proceed as described above and add `Iterable` in the future.

I think either way we'd still need it internally, so a type inferencer could conclude the constraint `x : Iterable[T]` upon seeing `for i in x`. Otherwise it'd have to conclude `x : Collection[T]`, which may be unusable for an application-defined type that has no definite size. If `Iterable` is not part of the user-visible type system, then the inferencer could narrow `Iterable[T]` to `Collection[T]` at the end of inferencing for the given function's body, in the absence of any specific information that `x : WeirdStreamType` instead.

brandjon:
> I just remembered https://github.com/bazelbuild/starlark/issues/203 - if we introduce views for dict.keys/items/etc., those views would be iterable but not collections.

Thanks! I forgot about that too.

`enumerate` and `zip` could in theory return the same type as they're given -- iterable, collection, or sequence -- though it might be awkward to specify that in a type signature since it wouldn't be parametric polymorphism. And anyway, Python doesn't allow taking the `len` or indexing of either `enumerate` or `zip`. So if we avoid using iterable for those builtins now, we risk being more permissive than Python and potentially wanting to claw back that expressivity in the future (`--incompatible_dont_let_users_take_len_of_enumerate`, etc.).

In Python, the `keys`, `values`, and `items` methods of `dict` can have their `len()` taken and support membership testing, but are not indexable. So they're still `Collection`s.

mahmoudimus:
> `Mapping` should subtype `Collection`, so that the `dict()` constructor can take a mapping and enumerate its keys (and query the associated values).

+1, I'm of the same opinion here. 

> I am wavering a little on whether we should drop `Iterable` or keep it in addition to `Collection`

So, are we thinking of keeping `Iterable`? I think it makes sense to have it. 

> Otherwise it'd have to conclude `x : Collection[T]`, which may be unusable for an application-defined type that has no definite size.

This is percisely how I've used it before. I've used it as a way to describe a lazy, deterministic operation for streams.

brandjon:
To confirm, my current position is that:

* We do `Sequence` <: `Collection` and `Mapping` <: `Collection` as described earlier
* We also now add in `Collection <: Iterable`
* All of these types are essentially compatible with their Python abc definitions, except that strings are nowhere in this hierarchy
* The various dict views are `Collections`s
* `range()` is a `Sequence`, but the other things like `enumerate()` and `zip()` are mere `Iterable`s
* If we ever add a `map()` function, it too returns an `Iterable`

One issue I didn't consider before is the behavior of the `in` operator. Python [specifies](https://docs.python.org/3/reference/expressions.html#membership-test-operations) that it'll fall back on iterating all elements if the rhs doesn't provide a `__contains__` method, so I guess we just specify that the rhs must be `Iterable` (not "sequence" as it currently says). This is consistent with keeping `enumerate()`/`zip()`/`map()` return types as iterable.

comius:
So if I follow the discussion correctly, the diff with current state of PR is to introduce also `Collection`?

brandjon:
Discussed further with Sasha: Let's leave off `Iterable` for now until we need it.

PR #320, opened 2025-08-12
Add recommendations for documenting Starlark APIs

tetromino:
Formalize the basics of what e.g. Stardoc expects from .bzl files.

Working towards https://github.com/bazelbuild/stardoc/issues/202

### DIFF
--- a/conventions.md
+++ b/conventions.md
@@ -70,7 +70,11 @@ DEFAULT_TAGS = {
 } #: Default list of tags for each configuration
 ```
 
+Documentation processing tools may treat multiple doc comments attached to the
+same variable as an error (for example, if a variable has both a preceding doc
+comment block and a trailing in-line doc comment).
+
 Doc comments attach to variables, not to values. For example, if a global
 variable whose value happens to be `True` has a doc comment, documentation
-processing tools should take care not to attach the doc comment's text to
-unrelated occurrences of `True` in other parts of the code.
+processing tools shouldn't attach the doc comment's text to unrelated
+occurrences of `True` in other parts of the code.


conventions.md
```diff
@@ -0,0 +1,68 @@
+# Starlark Language Conventions
+
+This document describes non-normative but generally recommended conventions for
+Starlark.
+
+## API Documentation
+
+### Docstrings
+
+API documentation for a function should be provided in a [Python-style
```
brandjon, line 10: Are we matching the PEP 257 standard exactly or describing a new standard merely inspired by it?

Are we saying anything about "Args:" sections, etc., or is that out of scope for the conventions?

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+docstring](https://peps.python.org/pep-0257/) - a string literal which is the
+first line of the function's body.
+
+By convention, this string should consist of an optional newline, followed by
```
brandjon, line 14: PEP 257 says that even one-line docstrings should use triple quotes. It also says one-line docstrings should not begin with a newline, i.e. ruling out

```
    def noop():
        """
        Do nothing.
        """
        pass
```

(Technically that's not considered a trivial multi-line docstring because it's not "followed by a more elaborate description" -- though it's possible I'm reading the PEP too literally.)

But the meta-comment here is that if we're going to say something about docstring style in the spec repo, then I'd like it to be clear exactly what is being prescribed.

At minimum, I'd say it's definitely in scope to mention the stuff that the parser specially recognizes -- docstrings as first statements of modules and function bodies, and doc comments before or inline with assignment statements. Anything beyond that we can either spell out here in precise detail, or incorporate by reference to a PEP or external resource, or leave unspecified and have Bazel- or Google-specific style guides fill in the gap.

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+delimited paragraphs of additional documentation text.
+
+If a string literal is the first statement in a Starlark source file, it is
+treated as the documentation for that file; the conventions are the same as for
+a function's docstring.
+
+For example:
+
+```python
+"""A collection of useful utilities"""
```
brandjon, line 25: Nit: According to the PEP it should end in a period. It also says function/method summary lines should be written in the imperative sense. I don't know that it makes sense to inherit either of those conventions here in the language spec repo. Maybe that's the answer to my earlier question of whether or not we're bound by PEP 257.

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Documentation processors should take care to dedent common leading whitespace
+from a multiline docstring's lines (note that the first line could have no
+leading whitespace).
+
+### Doc comments
+
+API documentation for a constant may be provided in
+[Sphinx-style](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings)
+*doc comments*, which start with `#:` optionally followed by a space.
```
brandjon, line 46: Nit: "optionally followed by a *single* space"

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+    ...
+```
+
+Documentation processors should take care to dedent common leading whitespace
+from a multiline docstring's lines (note that the first line could have no
+leading whitespace).
+
+### Doc comments
+
+API documentation for a constant may be provided in
```
brandjon, line 44: Not a "constant" (which I'd take to mean a lexical construct) -- a global variable.

But we can go further and say it's for any assignment statement. Or we can add on that it should only be for the first assignment statement, in the case of multiple assignments to the same local var. (Not sure what should happen if you violated that and tried to have two docs for the same local var.) Or we can leave it as is and have the implementation behavior of attaching doc strings to local assignments be an undocumented behavior.

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+[Sphinx-style](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings)
+*doc comments*, which start with `#:` optionally followed by a space.
+
+An uninterrupted block of doc comments attaches to the symbol on the left hand
+side of the immediately following assignment statement:
+
+```python
+#: List of allowed configuration names
+#: in an unspecified order
+ALLOWED_CONFIGS = ["foo", "bar"]
```
brandjon, line 54: What happens if it appears on a complex assignment? What does Sphinx do?

```
#: Coordinates of interest
x, y = get_coord()
```

conventions.md
```diff
@@ -0,0 +1,68 @@
...
+
+```python
+DEFAULT_TAGS = {
+    "foo": [],
+    "bar": ["local", "manual"],
+} #: Default list of tags for each configuration
+```
+
+If both the leading and trailing doc comment is specified, the trailing doc
+comment takes precedence.
```
brandjon, line 68: Is that Sphinx behavior? Seems error prone -- the special `#:` notation combined with the requirement that there be no intervening lines is a pretty clear signal that both comments are intended as documentation, so I don't see the utility of not failing fast.

I'd consider this to be in the same category as double-documenting the same variable at two different assignments (usually only possible in the local case, since standard Starlark dialect bans global reassignment).

conventions.md
tetromino: Sphinx applies the same doc string to both variables. The intended use case - I am guessing - is that only one of the variables is public and relevant:

```python
#: List of names ...
_, NAMES, _PRIVATE_DETAILS = foobar(...)
```

conventions.md
tetromino: Removed mention of PEP 257.

conventions.md
tetromino: Good point - I removed all mention of PEP 257.

conventions.md
tetromino: Added a period :)

conventions.md
tetromino: Fixed,

conventions.md
tetromino: Changed to global. Starlark doesn't allow reassigning a global, so the multiple assignment concern is moot :)

conventions.md
```diff
@@ -35,23 +31,28 @@ def is_valid(config):
...
-API documentation for a constant may be provided in
-[Sphinx-style](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings)
-*doc comments*, which start with `#:` optionally followed by a space.
+API documentation for a global variable may be provided in [Sphinx
+autodoc-style](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings)
+*doc comments*, which start with `#:` optionally followed by one space.
 
-An uninterrupted block of doc comments attaches to the symbol on the left hand
-side of the immediately following assignment statement:
+An uninterrupted block of doc comments attaches to the symbol(s) on the left
```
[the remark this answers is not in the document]
tetromino: Clarified text.

conventions.md
```diff
@@ -63,6 +64,3 @@ DEFAULT_TAGS = {
     "bar": ["local", "manual"],
 } #: Default list of tags for each configuration
 ```
-
-If both the leading and trailing doc comment is specified, the trailing doc
-comment takes precedence.
```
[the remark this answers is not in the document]
tetromino: Clarified. I don't think it's worth specifically mentioning multiple assignment - just say that a documentation processor is allowed to error out if it sees multiple doc comments on the same variable.

PR #321, opened 2025-09-05
Update prose on subscripting and slicing

brandjon:
* "Index expressions" are renamed to "Subscript expressions". This better matches Python's AST, and reflects the idea that subscripting is a syntactic operation (`a[b]`), whereas indexing is the special case where `a` is a sequence-like value and `b` is an integer.
* The explanation of indexing and slicing semantics is factored out from strings/bytes/lists/tuples, to instead reference the general sections on indexing and slicing.
* Drive-by: Elaborated on the semantics of `x in b` for bytes objects, and added the requirement (inherited from Python) that integer `x` is in range [0, 255).
* The section on indexing, which previously mentioned maps as an afterthought, is rewritten to be more general.
* Added explicit mention of the fact that Starlark doesn't support the `del` statement.
* Clarified the section on slicing.
* Clarified verbiage about complex assignments.

This is in the wake of #318 and on the way to related clarifications about sequence and mapping types.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -91,7 +91,7 @@ interact with the environment.
     * [Comprehensions](#comprehensions)
     * [Function and method calls](#function-and-method-calls)
     * [Dot expressions](#dot-expressions)
-    * [Index expressions](#index-expressions)
+    * [Subscript expressions](#subscript-expressions)
     * [Slice expressions](#slice-expressions)
     * [Lambda expressions](#lambda-expressions)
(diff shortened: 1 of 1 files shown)


tetromino (approved):

tetromino:
Thanks for this - it fixes a couple of wording inconsistencies about whether strings are sequences or not that were left after https://github.com/bazelbuild/starlark/pull/318

PR #323, opened 2025-09-09
Update prose and method signatures around iterable/collection

brandjon:
This follows up on #318 by making prose and builtin method type signatures consistent with the new collections terminology (iterable / collection / sequence / mapping).

* At long last, banish the phrase "iterable sequence", replacing it with either "iterable", `Collection`, or `Sequence`.

* Avoid some instances where non-`Sequence` types are informally referred to as "sequences", prefer the informal terms "container", "array", or even "series" where possible.

* In description of list data type etc., merge paragraphs about `Collection`/`Sequence`-related operations together.

* Rework "Collection types" section, break off the part on iteration to its own section and elaborate.

* Update many descriptions of built-in functions to specify the type as a collection or iterable. I generally tried to use iterable for all conversions/constructors, e.g. `list()` and `set()`. For other operations, I generally used collection in order to be more conservative unless/until we find we need to broaden it. Right now the difference is academic because I'm not aware of any non-Collection iterables in Bazel.

Minor changes:

* Data types section: streamline wording, remove "fixed-length" as it's not true for lists and redundant with "unmodifiable" for tuples.

* Mention you can't `+` lists with tuples.

* Consistently avoid hyphenation after "`for`" and "`if`".

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -103,7 +103,7 @@ interact with the environment.
     * [Return statements](#return-statements)
     * [Expression statements](#expression-statements)
     * [If statements](#if-statements)
-    * [For loops](#for-loops)
+    * [For loops](#for-statements)
     * [Break and Continue](#break-and-continue)
     * [Load statements](#load-statements)
   * [Module execution](#module-execution)
@@ -779,14 +779,14 @@ compared using operators such as `==` and `<`.
 (Beware that the UTF-16 string encoding is not order-preserving
 with respect to code point values.)
 
-Strings are _not_ iterable, so they cannot be used as the operand of
-a `for` loop, list comprehension, or any other operation than requires
-an iterable. Starlark deviates from Python here to avoid a common
-pitfall, in which a string is mistakenly used where a list of strings
-was intended, resulting in exploading the string into its individual
-characters. One must instead explicitly call a method of a string value
-to obtain an iterable view. Because strings are not iterable, they are not
-considered to be subtypes of `Collection` or `Sequence`.
+Strings are _not_ iterable, so they cannot be used as the operand of a `for`
+loop or of methods that expect iterables (such as the `list()` constructor).
+One must instead explicitly call a method of a string value to obtain an
+iterable view of its elements. Because strings are not iterable, they are not
+considered to be subtypes of `Collection` or `Sequence`. (Starlark deviates
+from Python here to avoid a common pitfall in which a single string is
+mistakenly used where a list of strings was intended, resulting in exploading
+the string into its individual characters.)
 
 Any value may formatted as a string using the `str` or `repr` built-in
 functions, the `str % tuple` operator, or the `str.format` method.
@@ -879,7 +879,7 @@ The [type](#type) of a list is `"list"`.
 Lists are a subtype of `Sequence`. The number of elements may be retrieved with
 the built-in `len()` function. Lists can be [indexed](#subscript-expressions)
 to retrieve a single element, and [sliced](#slice-expressions) to produce a new
-list. Lists are iterated over by `for` loops.
+list. Lists can be iterated over by `for` loops.
 
 List may be constructed using bracketed list notation:
 
@@ -1093,7 +1093,7 @@ A dictionary value has these methods:
 
 ### Sets
 
-A set is a mutable, collection of unique values - the set's *elements*.
+A set is a mutable collection of unique values - the set's *elements*.
 The [type](#type) of a set is `"set"`.
 
 Sets provide constant-time operations to insert, remove, or check for the
@@ -1461,8 +1461,8 @@ Except where noted, built-in functions accept only positional arguments.
 After a Starlark file is parsed, but before its execution begins, the
 Starlark interpreter checks statically that the program is well formed.
 For example, `break` and `continue` statements may appear only within
-a loop; `if`, `for`, and `return` statements may appear only within a
-function; and `load` statements may appear only outside any function.
+a `for` statement; `if`, `for`, and `return` statements may appear only
+within a function; and `load` statements may appear only outside any function.
 
 _Name resolution_ is the static checking process that
 resolves names to variable bindings.
@@ -1478,7 +1478,7 @@ Four Starlark constructs bind names, as illustrated in the example below:
 function parameters (`d`),
 and assignments (`e`, `h`, including the augmented assignment `e += h`).
 Variables may be assigned or re-assigned explicitly (`e`, `h`), or implicitly, as
-in a `for` loop (`f`) or comprehension (`g`, `i`).
+in a `for` statement (`f`) or comprehension `for` clause (`g`, `i`).
 
 ```python
 load("lib.star", "a", b="B")
@@ -1755,11 +1755,14 @@ function expects.
 <!-- Not referenced by name until we have type annotations, at which point they
 are type names. -->
 
-* `Collection`: A type that is *iterable* (can appear on the right-hand side
-  of a `for` loop), can have its length taken via the `len()` function, and
-  supports testing for membership via the `in` operator. Examples include
-  `list`, `tuple`, `set`, and `dict` (its elements are its keys), but not
-  `string` or `bytes` since they are not iterable.
+* `Collection`: A type that is:
+  * *iterable* (can appear on the right-hand side of a `for` loop, whether that
+    be a `for` statement or comprehension `for` clause),
+  * can have its length taken via the `len()` function, and
+  * supports testing for membership via the `in` operator.
+
+  Examples include `list`, `tuple`, `set`, and `dict` (its elements are its
+  keys), but not `string` or `bytes` since they are not iterable.
 
 * `Sequence`: A `Collection` that additionally supports subscript and slicing
   expressions, following the semantics of indexing. `list` and `tuple` are
@@ -2479,7 +2482,7 @@ and its result is a dictionary containing the key/value pairs
 for which the body expression was evaluated.
 Evaluation fails if the value of any key is unhashable.
 
-As with a `for` loop, the loop variables may exploit compound
+As with a `for` statement, the loop variables may exploit compound
 assignment:
 
 ```python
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -585,12 +579,12 @@ The truth value of `None` is `False`.
 There are two Boolean values, `True` and `False`, representing the
 truth or falsehood of a predicate.  The [type](#type) of a Boolean is `"bool"`.
 
-Boolean values are typically used as conditions in `if`-statements,
+Boolean values are typically used as conditions in `if` statements,
 although any Starlark value used as a condition is implicitly
 interpreted as a Boolean.
-For example, the values `None`, `0`, and the empty sequences
+For example, the values `None`, `0`, and the empty container values
```
tetromino: But `""` is not a container. I would suggest the following wording:

---

For example, the values `None`, `0`, the empty string `""`, and the empty
container values `()`, `[]`, and `{}` have a truth value of `False`, whereas
non-zero numbers, non-empty strings, and non-empty container values have a truth
value of `True`.

spec.md
```diff
@@ -785,10 +779,14 @@ compared using operators such as `==` and `<`.
 (Beware that the UTF-16 string encoding is not order-preserving
 with respect to code point values.)
 
-Strings are _not_ iterable sequences, so they cannot be used as the operand of
-a `for`-loop, list comprehension, or any other operation than requires
-an iterable sequence. One must explicitly call a method of a string value
-to obtain an iterable view.
+Strings are _not_ iterable, so they cannot be used as the operand of
```
tetromino: Minor copy-editing: extra comma, spelling fix, remove an "instead" that doesn't make sense in context of previous sentence:

----

Strings are *not* iterable, so they cannot be used as the operand of a `for`
loop, list comprehension, or any other operation than requires an iterable.
Starlark deviates from Python here to avoid a common pitfall in which a string
scalar value is mistakenly used where a list of strings was intended, resulting
in exploding the string into its individual characters. One must explicitly call
a method of a string value to obtain an iterable view of its elements. Because
strings are not iterable, they are not considered to be subtypes of `Collection`
or `Sequence`.

spec.md
```diff
@@ -877,9 +876,10 @@ TODO: string.elems(), string.elem_ords(), string.codepoint_ords()
 A list is a mutable sequence of values.
 The [type](#type) of a list is `"list"`.
 
-Lists are sequences: the elements of a list may be indexed by an integer,
-iterated over by `for`-loops, list comprehensions, and various built-in
-functions.
+Lists are a subtype of `Sequence`. The number of elements may be retrieved with
+the built-in `len()` function. Lists can be [indexed](#subscript-expressions)
+to retrieve a single element, and [sliced](#slice-expressions) to produce a new
+list. Lists are iterated over by `for` loops.
```
tetromino: Wording suggestion:

---

Lists can be iterated over by `for` loops and list comprehensions.

spec.md
```diff
@@ -1100,7 +1093,7 @@ A dictionary value has these methods:
 
 ### Sets
 
-A set is a mutable, iterable collection of unique values - the set's *elements*.
+A set is a mutable, collection of unique values - the set's *elements*.
```
tetromino: Typo: mutable collection (no comma)

spec.md
```diff
@@ -3050,9 +3058,10 @@ for a, i in [["a", 1], ["b", 2], ["c", 3]]:
   print(a, i)                          # prints "a 1", "b 2", "c 3"
 ```
 
-Because Starlark loops always iterate over a finite sequence, they are
+Because Starlark loops always iterate over a finite container (assuming
+the host application does not define an unbounded type), they are
 guaranteed to terminate, unlike loops in most languages which can
-execute an arbitrary and perhaps unbounded number of iterations.
+execute an arbitrary and perhaps unbounded number of iterations.)
```
tetromino: Remove extra `)` at end.

spec.md
```diff
@@ -3570,7 +3586,7 @@ reversed({"one": 1, "two": 2}.keys())           # ["two", "one"]
 ### set
 
 `set(x)` returns a new set containing the unique elements of the iterable
```
tetromino, line 3599: `x` is a collection, not an iterable; we don't support infinite sets

spec.md
```diff
@@ -3625,7 +3642,7 @@ UTF-K encoding of the replacement character, U+FFFD.
 
 ### tuple
 
-`tuple(x)` returns a tuple containing the elements of the iterable x.
+`tuple(x)` returns a tuple containing the elements of the iterable `x`.
```
tetromino, line 3656: Similarly to `set()`, here `x` must be a collection - we don't support infinite tuples.

tetromino (changes requested):

spec.md
brandjon: I asked the AI overlords (Gemini) and it came up with this:

>> In programming languages, what does the term "container" generally refer to?
>
> In programming, a container generally refers to a data structure that can store and manage a collection of other objects or elements. The primary purpose of a container is to provide an efficient way to organize, access, and manipulate data.
>
>> Would you consider a string to be a form of container?
>
> Yes, in most modern programming languages, a string is considered a form of container.
>
> A string meets the general definition of a container because it stores and manages a collection of elements, which in this case are individual characters.

It provides a link to a page from 2003 (lol) describing a "string container class" in C++. A search also turns up various hits referring to strings as containers in the context of C++ and Python.

spec.md
brandjon: Also "than" -> "that". ;)

Instead of replacing instead, I moved the comparison to Python to the end of the paragraph and made it a parenthetical. This avoids disrupting the flow.

spec.md
brandjon: I don't like having to say "comprehension" everywhere. (Note that it's not just list comprehensions but also dict comprehensions.)

I tried this instead: Everywhere we're talking about loops and comprehensions, I just say "`for` loop", and rely on the definition of a comprehension's `for` clause to imply that the same holds for that as well. There's one or two places where we need to specifically reference the statement not the clause, so for those I say "`for` statement". WDYT?

brandjon: Actually, I now mention the connection between "`for` loop" and the statement and clause in the definition of "iterable", appearing within the definition of `Collection`.

spec.md
brandjon: SGTM.

spec.md
brandjon: Done.

spec.md
brandjon: We don't support infinite lists either, but I have `list()` taking an iterable because I feel it's useful to allow these "constructor" builtins to be as broad as possible to support explicit conversions. For instance, `sorted(weird_builtin_iterable)` fails but `sorted(list(weird_builtin_iterable))` succeeds.

(Technically it'd be expressive enough if only a single collection's constructor accepted an iterable, but it seems strange to allow it for `list()` and not the other types.)

spec.md
brandjon: Nor infinite lists. :)

Note that a non-iterable collection is by no means necessarily infinite.

spec.md
tetromino: My brain apparently had mapped "container" to "collection" :) Since we don't use "container" anywhere else in the spec, I'd suggest for clarity

---

For example, the values `None`, `0`, the empty string `""`, and the empty
collection values `()`, `[]`, and `{}` have a truth value of `False`, whereas
non-zero numbers, non-empty strings, and non-empty collection values have a truth
value of `True`.

spec.md
tetromino: Fair enough. Note, however, that Bazel's set() implementation doesn't take iterables, and teaching it to take iterables would be a bit of a pain.

spec.md
tetromino: Agree with the new order. But typo: exploading should be exploding.

tetromino (approved):

spec.md
brandjon: In the current draft, the other uses of "container" are for explaining
* that we have no `del` statement, so you have to call a method on the container object instead
* that loops always iterate over a finite container and so are guaranteed to terminate

I've replaced those usages with "containing object" and "iterable" respectively.

For this paragraph, it's probably less verbose at this point (and more complete) to just use a list of bullet points. Updated. This also led me to clarify the surrounding prose and example.

spec.md
brandjon: Thanks.

spec.md
```diff
@@ -585,28 +579,35 @@ The truth value of `None` is `False`.
...
-numbers and non-empty sequences have a truth value of `True`.
+Boolean values are typically used as conditions in `if` statements,
+although any Starlark value may be used as a condition in this way.
+The following core language values are considered false when this happens:
+* `None`
+* numerical 0 (`0`, `0.0` and `-0.0`)
+* the empty string (`""`) and empty bytes (`b""`)
+* the empty collections `[]`, `()`, `{}`, and `set()`
+
+All other core langauge values are considered true.
```
tetromino: Nit: "language"

spec.md
```diff
@@ -585,28 +579,35 @@ The truth value of `None` is `False`.
...
-Boolean values are typically used as conditions in `if`-statements,
-although any Starlark value used as a condition is implicitly
-interpreted as a Boolean.
-For example, the values `None`, `0`, and the empty sequences
-`""`, `()`, `[]`, and `{}` have a truth value of `False`, whereas non-zero
-numbers and non-empty sequences have a truth value of `True`.
+Boolean values are typically used as conditions in `if` statements,
+although any Starlark value may be used as a condition in this way.
+The following core language values are considered false when this happens:
+* `None`
```
tetromino, line 586: Nit: should table include `False` itself, so as not to confuse the reader?

tetromino (approved):

spec.md
brandjon: I meant to refer strictly to non-Boolean values that get used as conditions, but I'll clarify.

spec.md
brandjon: Thanks!

PR #328, opened 2025-12-06
Fully define `min` and `max`

lgalfaso:
The definition of `min` and `max` was not taking into account the case that they were called with multiple positional arguments.

When called with one positional argument, clarify that the element must be iterable.

This extended definition is consistent with the current implementation of Bazel and with Python.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3461,7 +3461,13 @@ With no argument, `list()` returns a new empty list.
 
 ### max
 
-`max(x)` returns the greatest element in the collection `x`.
+`max(x)` returns the greatest element of the given arguments.
+
+Whenever called with one positional argument is provided, it must
+be an iterable and the largest item in the iterable is returned.
+
+Whenever called with two or more positional arguments, the
+largest of the positional arguments is returned.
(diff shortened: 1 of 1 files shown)


albertocavalcante:
@brandjon @tetromino can you please take a look? Thanks!

tetromino (changes requested):
Good catch that for min/max it's sufficient for the first argument to be an iterable, not necessarily a collection!

I would suggest changing the doc to make the method's signature clearer in the two cases. Maybe something like this:

----

`max` returns the greatest element.

When called with one positional argument, `max(x, key=None)` returns the greatest element in the iterable value `x`. It is an error if `x` is empty, not iterable, or if any of its elements do not support ordered comparison with others.

When called with 2 or more positional arguments, `max(key=None, *args)` returns the greatest of its positional `args`. It is an error if any of the positional arguments do not support ordered comparison with others.

It is an error to call `max` with no positional arguments.

lgalfaso:
Updated the PR with the recommended wording.

PR #334, opened 2026-02-01
Remove the case that dict·update can take `None`

lgalfaso:
None of the implementations accept `None` as a parameter for dict·update. This includes Bazel, starlark-go and Python. This construction is not accepted.


albertocavalcante:
@brandjon PTAL. Thanks!

PR #352, opened 2026-06-10
Add binary literals and underscore separators to number literals

### LINKED ISSUE #117: spec: 0b10101 binary literals
Are they supported? They work in C, C++, Python2,3, Go, Java, and Starlark/Go, and there's code for them in the Starlark/Java scanner, but they don't seem to actually work. They're also in the go.starlark.net language description but not the spec on this site.

We should add them to the spec and the Java implementation.

fmeum:
Extends the lexical grammar for number literals in two ways:

* **Binary integer literals** (`0b1011`/`0B1011`): these are documented in the go.starlark.net language description and partially implemented in the Java lexer, but were missing from the spec. Fixes #117.
* **Underscore digit separators** ([PEP 515](https://peps.python.org/pep-0515/)): a single `_` may appear between two digits of an int or float literal, or between a base prefix and the first digit (e.g. `1_000_000`, `0xFFFF_FFFF`, `0x_FF`, `1_000.000_1`). Underscores have no effect on the value.

Following Python, the spec also states that it is an error if a literal's digits are immediately followed by a decimal digit that is not valid in the literal's base (e.g. `0o12345678` is an error rather than the two tokens `0o1234567` and `8`).

This is the spec counterpart to the Java implementation in bazelbuild/bazel#29784, requested in [this review](https://github.com/bazelbuild/bazel/pull/29784#pullrequestreview-4462921255).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brandjon:
(CLA is false positive due to claude attribution)

spec.md
```diff
@@ -326,23 +329,36 @@ rb'hello'    br"hello"          # raw bytes literal
...
+binary_digit  = '0' | '1' .
 ```
 
+As in Python ([PEP 515](https://peps.python.org/pep-0515/)), a single
+underscore (`_`) may appear between two digits of a number literal, or
+between the base prefix and the first digit. Underscores serve only to
+group digits for readability and have no effect on the value of the
+literal.
+
+It is a static error if the digits of a number literal are immediately
```
brandjon, line 357: I would delete this second paragraph. I believe in all cases it is already a static error regardless of this text, so all the paragraph is doing is specifying that it's a lexical error rather than a parsing one, which is really an implementation detail.

brandjon (changes requested):

Issue #106, opened 2020-09-20
Add grammar to permit type annotations


ndmitchell:
Following up from https://github.com/bazelbuild/buildtools/issues/900, I think Starlark (or at least some extension thereof) should permit a grammar that allows inserting type annotations. Concretely (but informally) we should allow:

```
def myFunction(x : a_type) -> a_type:
```

Where `a_type` is any arbitrary expression. I think it would be useful to standardise where the types go, but not yet standardise what they mean or how they are evaluated. Some possibilities I can think of:

* Something Python3 compatible. Potentially allowing tools like MyPy or Pyre to analyse ot.
* Something more runtime checking, so they serve as assertions. I've experimented with this approach in a development branch, and while I can't share the code yet, it's been relatively effective.

Given that types restrict the meaning of a program, it should be perfectly compliant for "Standard Starlark" to parse but otherwise entirely ignore the type signatures. Starlark implementations could then have extensions that check (statically, as a lint, or at runtime) that the code is well-formed.

If there's general acceptance, I'm happy to put together a diff to the spec.

alandonovan:
The addition of types is an interesting idea---the lack of types is the biggest disadvantage of working in a Python-like language---but it is an enormous undertaking, and raises all kinds of questions, such as:
- Should dynamic checks be optional? If so, users can't rely on them for precondition checking, and nor can compiler optimizations.
- What is the performance cost of additional type checks, and can this be mitigated by compiler optimization?
- How would separate compilation work? Python, unlike typed languages, allows each file to be compiled without any knowledge of the values imported from other packages. Would that change? If so, it has enormous architectural ramifications for Bazel. If not, what limits does that place on the type system?
- To what extent have types been a success in Python3?  Which parts are good and which less so?
- Core Starlark has no struct type; that's a Bazelism. How would the type checker be made extensible to support application-defined data types with their own type-checking rules?
- Can types be added without compromising the core benefit of Starlark, namely that it is very simple?

I don't wish to dampen your enthusiasm, but I think it is very unlikely that the Starlark team will have significant time to spend advancing any type proposal, at least in the next year.

illicitonion:
This sounds really exciting, particularly being able to re-use existing tools like MyPy/Pyre.

In addition to the syntax proposed, some interesting cases to consider:
[...]

PR #362, opened 2026-08-27
Add type annotation syntax to the language spec

tetromino:
This is basically the syntax experimentally supported by Bazel at HEAD but with support for namespaced type names as used by Buck2.

It is also a subset of the Python3 type annotation syntax. (We want Starlark to be syntactically a subset of Python!)

This PR does not specify the semantics of type checking - it only specifies eval-time behavior of the new syntactic constructs.

If we at least agree on the syntax, it can be used by common tooling like buildifier et al.

Miscellaneous notes:

* `type` is not a keyword because we want to allow it to be used for variable names
* but `cast` and `isinstance` do have to be keywords because they take a type expression as argument, and type expressions *aren't* subsets of ordinary expression syntax (~~type applications allow trailing commas before the closing `]`, and~~ type syntax is allowed to use ellipsis `...` literals).

Working towards #106.

@brandjon @adonovan FYI

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -5113,7 +5113,6 @@ TypeArgument = TypeExpr
              | TypeDict
              | string
              | int
-             | '(' ')'
              | '...'
              .
 
--- a/spec.md
+++ b/spec.md
@@ -1904,14 +1904,13 @@ PrimaryExpr = Operand
             | PrimaryExpr DotSuffix
             | PrimaryExpr CallSuffix
             | PrimaryExpr SubscriptSuffix
-            | CastExpr
-            | IsInstanceExpr
             .
 
 Operand = identifier
         | int | float | string | bytes
         | ListExpr | ListComp
         | DictExpr | DictComp
+        | CastExpr | IsInstanceExpr
         | '(' [Expressions] [,] ')'
         .
 
@@ -2817,15 +2816,16 @@ in `for` statements and in comprehension `for` clauses.
 
 A typed assignment statement may be viewed as syntactic sugar for a [var
 statement](#var-statements) immediately followed by a simple assignment
-statement; it informs the type checker of a newly-bound variable's type and
-assigns the variable an initial value:
+statement; it informs the type checker of a new variable's type and assigns the
+variable an initial value:
 
 ```text
 AssignStmt = identifier ':' TypeExpr '=' Expressions.
 ```
 
-It is a static error for an assignment statement to (re)declare the type of a
-variable that has already been bound:
+A variable may only be associated with a type at the point in code where it is
+bound. Thus, a var statement or a typed assignment statement may only be used
+on a new binding:
 
 ```python
 def f():
@@ -3201,6 +3201,18 @@ A load statement within a function is a static error.
 
 ## Type annotations
 
+Type annotations are used to inform static type checkers for Starlark. The
+semantics of type checking are not specified by this document, but are intended
+to be broadly similar to those of Mypy.
+
+Identifiers in type expressions are resolved statically, following the same
+rules as in value expressions.
+
+Type annotations described below are evaluated statically. However, most code
+(with the exception of the `...` ellipsis token) that may be parsed as a type
+expression may also be parsed as a value expression, and then evaluated
+dynamically.
+
 ### Type expressions
 
 Type expressions consist of one or more *type atoms* separated by the `|`
@@ -3215,14 +3227,22 @@ TypeExpr = TypeAtom
 Semantically, `|` in a type expression means "or" (a union type). For example,
 `int | float | None` denotes "integer or float or `None`".
 
-Type atoms can be plain *type names* (an identifier or a sequence of identifiers
-separated by `.` for namespacing), or *type applications* with type arguments in
-square brackets:
+A type atom can be a plain *type name* (an identifier or a sequence of
+identifiers separated by `.` for namespacing), a *type application* with type
+arguments in square brackets, or a *parenthesized tuple* of type expressions
+(or a single parenthesized type expression).
+
+The rules for parsing parenthesized type expressions and parenthesized tuples
+are identical to those for parsing parenthesized tuples and parenthesized
+expressions in value syntax.
 
 ```
 TypeAtom = TypeName
          | TypeApplication
+         | '(' [TypeExprs [',']] ')'
          .
+
+TypeExprs = TypeExpr {',' TypeExpr}.
 
 TypeName = identifier
          | TypeName DotSuffix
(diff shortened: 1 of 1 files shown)


brandjon:
Discussed offline: I believe we should make type expressions subset value expressions (i.e. type applications are subscript expressions) by eliminating tolerance for trailing commas in type applications except where the type application arguments are wrapped in an extra tuple (as in `dict[(K, V,)]`). We can also at some point relax the syntactic prohibition on ellipses appearing in value expressions, although it has no real use and can still be a dynamic error.

Reasoning:
* We want buildifier and possibly other tools to have an easier time handling expressions that may be either type expressions or value expressions (since the context is ambiguous if you are an older tool that doesn't know about `cast`/`isinstance`).
* We (or at least I) don't want to tolerate trailing commas for unparenthesized singleton tuples, as that's a recipe for subtle mistakes.
* Tolerating trailing commas for unparenthesized non-singleton tuples leads to an inconvenience when editing a multiline tuple from >1 elements to only 1 element.
* Although intolerance of trailing commas is annoying, it's precedented in Starlark, and IMO preferable to extending value syntax.

brandjon:
It's worth noting that in all of google3 I wasn't able to find any instances of `isinstance` or `cast` appearing as a user-defined identifier, although it did occur in comments (SQL or complaints about Starlark not supporting `isinstance()`).

spec.md
```diff
@@ -1903,6 +1904,8 @@ PrimaryExpr = Operand
             | PrimaryExpr DotSuffix
             | PrimaryExpr CallSuffix
             | PrimaryExpr SubscriptSuffix
+            | CastExpr
```
brandjon, line 1907: What about factoring the two new type-based expressions to `Operand`? Seems to me `PrimaryExpr` is more about permitting a `*Suffix` nonterminal, while the more atomic and easily parsed productions go into `Operand`.

spec.md
```diff
@@ -2810,6 +2813,25 @@ a, b = 2, 3
...
+statement](#var-statements) immediately followed by a simple assignment
+statement; it informs the type checker of a newly-bound variable's type and
+assigns the variable an initial value:
+
+```text
+AssignStmt = identifier ':' TypeExpr '=' Expressions.
+```
+
+It is a static error for an assignment statement to (re)declare the type of a
+variable that has already been bound:
```
brandjon, line 2828: "already been bound by a lexically preceding piece of syntax."

Or better yet, something like "A variable may only be associated with a type at the point in code where it is bound. Thus, a var statement or a typed assignment statement may only be used on a new binding."

tetromino:
Agreed. I've removed trailing commas from type application syntax.

tetromino:
> We can also at some point relax the syntactic prohibition on ellipses appearing in value expressions, although it has no real use and can still be a dynamic error.

Actually, there is a strong use case for unifying type/value syntax and allowing ellipses in value expressions: provider fields. As far as I can tell, Buck2's `provider_field()` (see https://buck2.build/docs/api/build/#provider_field) is takes the type as a value expression.

And in Bazel we will run into exactly the same use case. Consider annotating field types in something like

```bzl
MyInfo = provider(fields = ["foo"])
```

We want to say that `foo` is a field of type `struct[{"x": str}, ...]` (with an ellipsis). How do we do it? How do we convert type syntax into a reified value? Some possibilities within the bounds of the current syntax are

1. dedicated keyword with a call-like syntax (something like `MyInfo = provider(fields = {"name": provider_field(struct[{"x": str}, ...])`) }), like we have dedicated keywords for `cast` and `isinstance`;
2. declaring the type in a LHS annotation (something like `MyInfo: Provider[{"name": struct[{"x": str}, ...]}] = provider(<...>)`);
3. string parsing (something like `MyInfo = provider(fields = {"name": "(struct[{'x': str}, ...]): doc doc doc..."})`)

... but all 3 of those are bad. Option 1 is bad because the least unnatural keyword to use (`provider_field`, to match Buck2) would have to be elevated to *syntax level*, which is completely inappropriate for an application-specific concept. Option 2 duplicates field names on two sides of the assignment. Option 3 is very error-prone.

*By far* the most natural option is to support type-land operation at evaluation time for reified type values, so that the reified type `struct` allows an eval-time subscripting operation that aplies the appropriate type constructor to produce another reified type, etc. But this means supporting ellipsis in value syntax!

tetromino:
As pointed out by @dtolnay in https://github.com/bazel-contrib/buildtools/issues/1508, Buck2 uses tuple literal syntax for fixed-length tuple types: `()` is the empty tuple type, `(T,)` is a 1-element tuple, `(T,U)` (or `(T,U,)`) is a 2-element tuple, etc. And moreover, Buck2 allows parenthesized type expressions.

We'll need to support such syntax as type atoms.

brandjon:
Hm? You're saying Buck 2 uses `(A, B, C)` as an alias for `tuple[A, B, C]`, and that we want compatibility with Buck 2's way of writing tuple types?

I think we already support parenthesized type expressions, do we not?

spec.md
```diff
@@ -2925,6 +2954,16 @@ This is called the _keyword arguments_ parameter, and accumulates in a
 dictionary any surplus `name=value` arguments that do not match a
 prior parameter. It is conventionally named `**kwargs`.
 
+The name of any parameter in the parameter list except for the bare `*` may
+optionally be followed by a type annotation (a `:` followed by a [type
+expression](#type-expressions)). For ordinary parameters (required, optional, or
+keyword-only), this type expression denotes the type of the parameter itself.
+For the `*args` and `**kwargs` parameters, the type annotations denote the type
+of individual values accumulated by `*args` or `**kwargs` respectively.
```
brandjon, line 2962: I know Ivo didn't like Python's precedent for *args/**kwargs. I'm not necessarily a fan of how Python does it but I think it's probably more valuable to keep it than invent our own thing with the same apparent syntax. But I'm curious if you know what Buck does?

spec.md
```diff
@@ -2925,6 +2954,16 @@ This is called the _keyword arguments_ parameter, and accumulates in a
...
+
+Finally, between the parameter list and the `:`, the function may optionally
+annotate its return type by an *arrow* `->` followed by a type expression.
+
```
brandjon, line 2966: Do we have anything saying what the *semantics* of type expressions are with respect to the runtime environment? We need at minimum to say that:

* Identifiers in type expressions are resolved statically, following the same rules as in value expressions. TypeParams create new bindings whose scope is the params list, return type expression, and body.

* No type expression is evaluated dynamically, although some (or all) code that may be parsed as a type expression can also be parsed as a value expression.

spec.md
tetromino: Same as us: https://github.com/facebook/starlark-rust/blob/main/starlark/src/typing/tests/call.rs

spec.md
tetromino: Good point. Added.

spec.md
tetromino: Good idea.

spec.md
tetromino: Done.

tetromino:
Updated with suggestions + support for Buck2-style typles/parenthesized exprs.

@brandjon, should I try to tackle `...` as a value expr in this PR, or as a follow-up? (My preference would be to do it in a follow-up.)

